PR #3, opened 2024-04-30
Wes/vite to webpack

wesrisenmay-mx:


src/index.css
tphalp, line 1: Do we need both .css files?

tphalp (approved):
Just one comment about the css files.

PR #10, opened 2024-05-13
added workspaces for the institution and authentication services

wesrisenmay-mx:


tphalp (approved):
Thanks for adding these.

PR #18, opened 2024-06-21
Wes/registration login

wesrisenmay-mx:


.github/workflows/ui-e2e-tests.yml
```diff
@@ -18,15 +18,18 @@ jobs:
...
         uses: cypress-io/github-action@v6
+        env:
+          CYPRESS_auth_username: ${{ secrets.AUTH_USERNAME }}
+          CYPRESS_auth_password: ${{ secrets.AUTH_PASSWORD }}
 
         with:
           project: apps/ucp-ui
-          start: npm run dev
+          start: npm run dev --workspace=apps/ucp-ui
           wait-on: "http://localhost:8080"
```
tphalp, line 28: What do you think about using a different port range, so it's less confusing with the services that are already running in the 8xxx range. 3xxx is a common ui range.

tphalp (approved):

.github/workflows/ui-e2e-tests.yml
wesrisenmay-mx: Done

tphalp (approved):

PR #22, opened 2024-07-29
Add secret rotation code to Auth Service

tphalp:
Adds necessary code, including tests, for secret rotation.

Resolves UCP-172

### DIFF
--- a/apps/authentication-service/src/resources/clients/clientsRoutesHandlersV1.ts
+++ b/apps/authentication-service/src/resources/clients/clientsRoutesHandlersV1.ts
@@ -99,7 +99,8 @@ export const clientsRotateSecretsV1 = async (req: Request, res: Response) => {
 
     res.send(
       JSON.stringify({
-        ...client,
+        clientId: client.client_id,
+        clientSecret: client.client_secret,
       }),
     );
   } catch (error) {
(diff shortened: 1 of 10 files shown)


apps/authentication-service/src/resources/clients/clientsRoutesHandlersV1.ts
```diff
@@ -89,3 +90,30 @@ export const clientsDeleteV1 = async (req: Request, res: Response) => {
...
+    ) {
+      res.status(404).json({ message: "Client not found" });
+    } else if (
+      error instanceof Error &&
+      error.message.toUpperCase().includes("TOO MANY REQUESTS")
+    ) {
+      res.status(429).json({ message: "Rate limit exceeded" });
+    } else {
+      res.status(500).json({ message: "Unable to rotate client secret" });
+    }
```
Craiting, line 117: Instead of duplicating this code I would recommend creating a method which can be used in all places where this is duplicated.

tphalp: Okay, I refactored these checks into a separate method.

tphalp: @Craiting FYI ^

Craiting (approved):

PR #24, opened 2024-07-31
Adding rate limiter to auth service

tphalp:
This adds a rate limiter so our APIs won't have the potential to be overloaded.


apps/authentication-service/cypress/e2e/express.cy.ts
```diff
@@ -18,4 +18,23 @@ describe("Express", () => {
...
+      } else {
+        cy.request({
+          url: `http://localhost:${PORT}/ping`,
+          failOnStatusCode: false,
+        })
+          .its("status")
+          .should("eq", 429);
+      }
+    }
+  });
```
tphalp, line 39: I'm not sure if this test is useful. It does seem to test the rate-limiter live, but it might be brittle. It relies on very specific numbers in order to work, which I guess isn't that big of a deal.

I'd love feedback on a better way to test this. I couldn't find anything useful online.

apps/authentication-service/package.json
```diff
@@ -42,9 +43,9 @@
...
     "eslint-plugin-prettier": "^5.1.3",
     "jest": "^29.7.0",
-    "msw": "^2.3.0",
+    "msw": "^2.3.4",
     "prettier": "^3.2.5",
-    "ts-jest": "^29.1.2",
+    "ts-jest": "^29.1.5",
     "ts-node": "^10.9.2",
     "tsconfig-paths": "^4.2.0",
     "typescript": "^5.4.5"
```
tphalp, line 51: Updated dep versions

apps/authentication-service/src/init.test.ts
```diff
@@ -8,76 +8,83 @@ import { errorHandler } from "@/middleware/errorMiddleware";
 import { notFoundHandler } from "@/middleware/notFoundMiddleware";
 
 describe("Express test", () => {
-  it("tests notFoundHandler", () => {
-    const res = {
-      status: jest.fn().mockReturnThis(),
-      json: jest.fn(),
-    } as unknown as Response;
+  describe("middleware tests", () => {
```
tphalp, line 11: I just added more descriptive test names.

apps/authentication-service/src/init.ts
```diff
@@ -1,10 +1,16 @@
...
+import { rateLimit } from "express-rate-limit";
+import express, { Application, NextFunction, Request, Response } from "express";
+
 import { errorHandler } from "@/middleware/errorMiddleware";
 import { notFoundHandler } from "@/middleware/notFoundMiddleware";
+import { clientsRoutes } from "@/resources/clients/clientsRoutes";
+
+const rateLimitWindowMinutes = 10;
+
+const isTesting = () => process.env.ENV === "test" || process.env.IS_GITHUB;
```
tphalp, line 13: Added method to tell whether we're in testing mode or not. This was especially necessary for GH CI/CD to work. 

The ENV var on GH is set to `prod`, which seems necessary for some tests, because, if I changed it to `test`, some of them broke. 

So, I added a GH VAR so I could detect when it's running on GH.

apps/authentication-service/src/init.ts
```diff
@@ -16,6 +22,17 @@ export const initExpress = (app: Application): void => {
...
+    windowMs: isTesting() ? 2 * 1000 : rateLimitWindowMinutes * 60 * 1000, // (2 seconds for testing) or 10 minutes for production
+    limit: isTesting() ? 5 : 500, // max average 5/500 requests per windowMs (2 seconds/10 minutes)
+    message: `Too many requests from this IP, please try again after ${rateLimitWindowMinutes} minutes`,
+    handler: (_req, res, _next, options) =>
+      res
+        .status(options.statusCode)
+        .json({ message: options.message as string }),
+  });
+  app.use(limiter);
+
```
tphalp, line 35: Rate limit middleware added.

apps/authentication-service/cypress/e2e/express.cy.ts
Craiting: From what I can gather from the test the rate limit gets hit after 3 api requests in some small amount of time. I don't feel like the exact numbers are very important to test or have defined in a test. You could make the test description a little better like: "rate limit is reached after many quick requests in under <some amount of time>"

Then probably a safer test would probably be something along the lines of 
```javascript
expect {
  for (let i = 0; i < 10; i++) {
    cy.request({url: ...})
}.to throw an error with 429 status
```

I think you could do something like that, if you don't use `failOnStatusCode` maybe it will throw the error.

tphalp:
Looks like the rate limiter is causing the other cypress tests to fail. Looking into it.

apps/authentication-service/src/init.ts
Craiting: Interesting, in ucw-app the env var is set to test. Doesn't it make sense to have the github environment to be "test" since it's literally a test environment?

Craiting (commented):
I left an idea I had for solving your async problem. Let me know what you think.

.github/workflows/auth-e2e-service-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+          echo PORT="8089" >> ${ENV_FILE_PATH}
+          
+          # Secrets (can't load these from another job, due to GH security features)
+          echo AUTH0_CLIENT_SECRET=${{ secrets.AUTH0_CLIENT_SECRET }} >> ${ENV_FILE_PATH}          
+          echo E2E_CLIENT_SECRET=${{ secrets.E2E_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
+          echo E2E_PASSWORD=${{ secrets.E2E_PASSWORD }} >> ${ENV_FILE_PATH}
+
+          cat ${ENV_FILE_PATH}
+
+      - name: Cypress run
```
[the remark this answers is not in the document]
tphalp: Oh, I like this. Thank you. So, I would go back to having one cypress workflow file, but just add another step, so there's one specifically for ratelimiting, and another for the other tests. And then I split the rate-limiting file into a different folder?

apps/authentication-service/src/init.ts
tphalp: I've been thinking about this. I could see having GH set to "prod", since we want the tests to mimic prod, but right now, at least in the auth service for the rate-limiting test, the different ENV values don't change anything.

Craiting (approved):
Looks great thanks for trying that idea

PR #26, opened 2024-08-02
Ct/load institution db

Craiting:
Loading the institution json file into the database.

Also adds `institution_provider_id` because it's an essential field that was not included by mistake.

### DIFF
--- a/apps/institution-service/tasks/loadInstitutionsFromJson.ts
+++ b/apps/institution-service/tasks/loadInstitutionsFromJson.ts
@@ -1,0 +1,168 @@
+import fs from "fs";
+import path from "path";
+import db from "../config/database";
+import { Institution } from "../models/institution";
+import { Provider } from "../models/provider";
+
+export interface CachedInstitution {
+  name: string;
+  keywords: string | null;
+  logo: string;
+  url: string;
+  ucp_id: string;
+  is_test_bank: boolean | false;
+  routing_numbers: string[];
+  mx: InstitutionProvider;
+  sophtron: InstitutionProvider;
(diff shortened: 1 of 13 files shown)


apps/institution-service/tasks/loadInstitutionsFromJson.ts
```diff
@@ -0,0 +1,114 @@
...
+import { Institution } from "../models/institution";
+import { Provider } from "../models/provider";
+
+export interface CachedInstitution {
+  name: string;
+  keywords: string | null;
+  logo: string;
+  url: string;
+  ucp_id: string;
+  is_test_bank: boolean | false;
```
tphalp, line 13: I've never seen `boolean | false` before. What does this do?

tphalp: I was thinking maybe it's a way to set a default value, but I can't find any documentation to support that. I don't think using `boolean | false` is necessary here. `boolean` should suffice, unless I'm not understanding the meaning here.

tphalp (commented):
I left a comment.

apps/institution-service/tasks/loadInstitutionsFromJson.ts
Craiting: Yeah that was an accident, it was supposed to be boolean | null. I'll fix it.

Craiting: Actually I don't even need the `| null` at all anymore.

tphalp (approved):

PR #27, opened 2024-08-02
Add cypress to institution service

tphalp:
- Add cypress module, along with needed folder structure, config, npm run commands, etc.
- Adds e2e test for 'ping' endpoint
- Remove need for dotenv, and hard-code PORT


apps/institution-service/cypress/support/commands.ts
```diff
@@ -0,0 +1,46 @@
+/// <reference types="cypress" />
+import "@testing-library/cypress/add-commands";
```
wesrisenmay-mx, line 2: We need the following:

1. github action that runs this test suite
2. At minimum a fake cypress test, or add a ping endpoint and ping that
3. package.json commands to open and run cypress

tphalp: I'll add 1 and 2. For 3, these commands already existed in the package file, so weren't necessary to add in this PR.

apps/institution-service/cypress.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "cypress";
+import "dotenv/config";
+
+export default defineConfig({
+  env: {
+    PORT: process.env.PORT || 8088,
```
wesrisenmay-mx, line 6: Why don't we just hardcode a port?

tphalp: Because I took this from the auth service, and that's how it was done there.

tphalp: Oh, actually, that's not true. I'll hard-code it.

.github/workflows/institution-e2e-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+      - run: npm ci
+
+      - name: 'Create env file'
+        run: |
+          ENV_FILE_PATH=./apps/institution-service/.env
+          touch ${ENV_FILE_PATH}
+          
+          # Vars
+          echo -e "${{ needs.setup-env.outputs.env_vars }}" >> ${ENV_FILE_PATH}
+          echo PORT="8088" >> ${ENV_FILE_PATH}
```
wesrisenmay-mx, line 33: Need to remove this. Probably in the auth service action too

tphalp: Let's talk about how we decide on port numbers. The auth service uses .env for this. We can do static, but we should decide the best way to proceed. Do they need to be different, or does that not matter, etc.

.github/workflows/institution-e2e-tests.yml
```diff
@@ -0,0 +1,38 @@
+name: Institution E2E
+
+on: pull_request
+
+jobs:
+  setup-env:
```
wesrisenmay-mx, line 6: We don't need this anymore right?

apps/institution-service/app.ts
```diff
@@ -1,11 +1,9 @@
-import dotenv from "dotenv";
 import express, { Request, Response } from "express";
 import logger from "morgan";
 import institutionRoutes from "./routes/institutionRoutes";
 
-dotenv.config();
 const app = express();
-const PORT = process.env.PORT || 3030;
+const PORT = process.env.PORT || 8088;
```
wesrisenmay-mx, line 6: I'd rather it fails if it's misconfigured

wesrisenmay-mx (approved):

PR #28, opened 2024-08-06
removing institution crud until we're ready to use it

wesrisenmay-mx:


tphalp (approved):
So sad removing all those tests. 😭 😁

PR #29, opened 2024-08-06
Add role and permission based middleware

tphalp:


.github/workflows/auth-e2e-tests.yml
```diff
@@ -44,13 +44,18 @@ jobs:
         with:
           project: apps/authentication-service
           start: npm run dev --workspace apps/authentication-service
-          wait-on: "http://localhost:${{ vars.PORT }}/ping"
+          wait-on: "http://localhost:8089/ping"
```
tphalp, line 47: Wes and I talked about this. There's no need to have PORT in an .env file, since only UCP will be using this.

apps/authentication-service/cypress/e2e/rateLimiting.cy.ts
```diff
@@ -1,4 +1,4 @@
-const testLimit: number = 30;
+const testLimit: number = 40;
```
tphalp, line 1: Increased this to accommodate additional e2e tests. Is there a better way to do this? Since tests can happen in parallel, they all add to the rate limit max.

apps/authentication-service/cypress.example.env.json
```diff
@@ -2,5 +2,7 @@
   "E2E_CLIENT_ID": "osS8CuafkPsJlfz5mfKRgYH942Pmwpxd",
   "E2E_CLIENT_SECRET": "",
   "E2E_USERNAME": "tyson.phalp+cypress@mx.com",
-  "E2E_PASSWORD": ""
+  "E2E_PASSWORD": "",
+  "E2E_USERNAME_BASIC": "tyson.phalp+cypressbasic@mx.com",
```
wesrisenmay-mx, line 6: What does basic mean?

apps/authentication-service/cypress/support/commands.ts
```diff
@@ -39,29 +40,36 @@ import { JwtPayload } from "jsonwebtoken";
 //   }
 // }
 
-Cypress.Commands.add("loginClientAuth0", () => {
+Cypress.Commands.add("loginClientAuth0", (doUseRoleUser: boolean = true) => {
```
wesrisenmay-mx, line 43: What role are we talking about?

wesrisenmay-mx, line 43: I'd use should instead of do

wesrisenmay-mx, line 43: This abstraction doesn't feel very extensible. I'd do something more like this: https://github.com/Universal-Connect-Project/ucp-hosted-app/blob/main/apps/ucp-ui/cypress/support/commands.ts

apps/authentication-service/cypress.example.env.json
wesrisenmay-mx: I think these users need some more obvious names about what makes them different

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -161,4 +161,60 @@ describe("Client API", () => {
       expect(response.body).property("message").to.eq("Keys not found");
     });
   });
+
+  it("tries CREATING a client with a token that does not have the proper permissions", () => {
```
wesrisenmay-mx, line 165: The it statement should have more clarity on what the expected outcome is.

it("responds with a 403 when creating api keys without the proper permissions")

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -161,4 +161,60 @@ describe("Client API", () => {
...
+    }).then((response: Cypress.Response<{ body: Keys }>) => {
+      expect(response.status).to.eq(403);
+    });
+  });
+
+  it("tries GETTING a client with a token that does not have the proper permissions", () => {
+    cy.request({
+      failOnStatusCode: false,
+      method: "GET",
+      url: `http://localhost:${PORT}/v1/clients/keys`,
```
wesrisenmay-mx, line 183: This url is used a lot. Time for a variable.

apps/authentication-service/src/init.test.ts
```diff
@@ -61,7 +61,9 @@ describe("Express test", () => {
 
       // eslint-disable-next-line @typescript-eslint/unbound-method
       expect(res.status).toHaveBeenCalledWith(401);
-      expect(res.json).toHaveBeenCalledWith({ message: "Invalid Token" });
+      expect(res.json).toHaveBeenCalledWith({
```
wesrisenmay-mx, line 64: How was this passing before you made this change?

apps/authentication-service/src/middleware/errorMiddleware.ts
```diff
@@ -13,13 +13,13 @@ export const errorHandler = (
   let message: string = "Internal Server Error";
 
   if (error instanceof InvalidTokenError) {
```
wesrisenmay-mx, line 15: The outcome of these if statements is identical. Should they be merged?

apps/authentication-service/src/test/testData/users.ts
```diff
@@ -67,20 +68,10 @@ export const exampleUserAlreadyHasAClientError = {
...
-      e: "AQAB",
-      kid: "0",
-    },
-  ],
-};
-
-export const getTestToken = (isExpired = false): string => {
+export const getTestToken = (
+  isExpired: boolean = false,
+  doAddWidgetRolePermissions: boolean = true,
```
wesrisenmay-mx, line 73: should instead of do

apps/authentication-service/src/test/testData/users.ts
```diff
@@ -113,18 +104,27 @@ export const getTestToken = (isExpired = false): string => {
...
     header: {
       alg: "RS256" as Algorithm,
       kid: "0",
     },
     algorithm: "RS256" as Algorithm,
     audience: [
-      `https://${envs.AUTH0_AUDIENCE}/api/v2/`,
-      `https://${envs.AUTH0_AUDIENCE}/userinfo`,
+      doAddWidgetRolePermissions ? envs.AUTH0_CLIENT_AUDIENCE : "",
+      doAddWidgetRolePermissions ? "" : `https://${envs.AUTH0_DOMAIN}/api/v2/`,
```
wesrisenmay-mx, line 126: Are 126 and 127 necessary for tests to work?

apps/ucp-ui/src/App.tsx
```diff
@@ -8,16 +8,23 @@ import "@fontsource/work-sans/500.css";
 import "@fontsource/work-sans/700.css";
 import { Experimental_CssVarsProvider as CssVarsProvider } from "@mui/material";
 import { muiTheme } from "./theme";
+import {
+  DefaultPermissions,
+  WidgetHostPermissions,
+} from "./shared/constants/roles";
 
 const App: React.FC = () => {
+  const scope = `${Object.values(DefaultPermissions).join(" ")} ${Object.values(WidgetHostPermissions).join(" ")}`;
```
wesrisenmay-mx, line 17: How does scope work if you don't have the necessary permissions?

tphalp: You will get the token, but it won't have access to the endpoints that required certain scopes.

apps/authentication-service/src/test/testData/users.ts
wesrisenmay-mx: Let's nuke them then

apps/authentication-service/cypress/support/commands.ts
tphalp: I got rid of this by refactoring the login code, as you suggested.

tphalp:
@wesrisenmay-mx I believe I've addressed all of your feedback. Thank you for that.

wesrisenmay-mx (approved):
Nice work

PR #30, opened 2024-08-07
Ct/get cached institution list

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-58

Sets up the endpoint to get all institutions in the cachedInstitution format.


apps/institution-service/.env.example
```diff
@@ -0,0 +1,10 @@
+AUTH0_AUDIENCE=
+AUTH0_DOMAIN=
+
+# Cypress testing variables
+E2E_CLIENT_ID=
```
wesrisenmay-mx, line 5: We should keep cypress out of our main env files. 

Example: https://github.com/Universal-Connect-Project/ucp-hosted-app/blob/main/apps/ucp-ui/cypress.example.env.json

apps/institution-service/controllers/institutionController.ts
```diff
@@ -9,4 +10,35 @@ export const getAllInstitutions = async (req: Request, res: Response) => {
...
+
+    const cachedInstitutionList = institutions.map(
+      transformInstitutionToCachedInstitution
+    );
+
+    res.status(200);
+    res.json(cachedInstitutionList);
+  } catch (error) {
+    console.log("error", error);
+    res.status(400).json({ error: "Error getting all Institutions" });
```
wesrisenmay-mx, line 42: Need tests for this

apps/institution-service/cypress/e2e/express.cy.ts
```diff
@@ -1,12 +1,23 @@
-import { PORT } from "./../../shared/const";
-
 describe("Express", () => {
   it("returns pong", () => {
-    cy.request(`http://localhost:${PORT}/ping`).then(
+    cy.request(`http://localhost:8088/ping`).then(
```
wesrisenmay-mx, line 3: Why this change?

apps/institution-service/openApi.json
```diff
@@ -0,0 +1,120 @@
...
+      "bearerAuth": {
+        "type": "http",
+        "scheme": "bearer",
+        "bearerFormat": "JWT"
+      }
+    }
+  },
+  "servers": [
+    {
+      "description": "Local development widget",
```
wesrisenmay-mx, line 19: These 2 lines need updated

apps/institution-service/.env.example
```diff
@@ -0,0 +1,10 @@
+AUTH0_AUDIENCE=
```
wesrisenmay-mx, line 1: I don't think these 2 variables are sensitive, and they're required for this service to function

wesrisenmay-mx: I think we should move them out of the env and hardcode them in a variable somewhere

apps/institution-service/.env.example
Craiting: I saw that and thought it was kinda confusing because you look at the [cypress.config.ts](https://github.com/Universal-Connect-Project/ucp-hosted-app/blob/89cabb8f1799da3fca0f2bd67326b5edb86c96df/apps/institution-service/cypress.config.ts#L6-L11) and it looks like all the env variables are coming from the .env file. And then locally you have a cypress.example.env which overwrites those variables. But on the github that file doesn't exist so it just pulls the variables from the environment variables.

apps/institution-service/cypress/e2e/express.cy.ts
Craiting: Idk why I changed that, I'll revert it

apps/institution-service/.env.example
Craiting: Are you trying to say it should be hardcoded everywhere it's used? I could see these variables potentially being different for different environments so it makes sense to me to have them here.

apps/institution-service/cypress/support/e2e.ts
```diff
@@ -14,4 +14,46 @@
...
+      client_id,
+      client_secret,
+    },
+  }).then((response: Cypress.Response<JwtPayload>) => {
+    Cypress.env("ACCESS_TOKEN", response.body.access_token);
+  });
+
+  const username = Cypress.env("E2E_USERNAME") as string;
+  const password = Cypress.env("E2E_PASSWORD") as string;
+  const noAccessClientId = Cypress.env("E2E_CLIENT_ID") as string;
```
wesrisenmay-mx, line 40: Is this the web ui client id and client secret? If so, then we should call it that

apps/institution-service/cypress/support/e2e.ts
```diff
@@ -14,4 +14,46 @@
...
+  const noAccessClientSecret = Cypress.env("E2E_CLIENT_SECRET") as string;
+  const userAudience = Cypress.env("AUTH0_AUDIENCE") as string;
```
wesrisenmay-mx, line 42: Does this fail if you use the auth0_widget_audience?

apps/institution-service/cypress/support/e2e.ts
Craiting: I'm not sure but I can figure it out. Do you want the env variable to be renamed or just the local variable?

Craiting: Just checked and yes it is the UCP Web UI Client

apps/institution-service/cypress/support/e2e.ts
Craiting: Yeah

apps/institution-service/cypress/support/e2e.ts
wesrisenmay-mx: Yeah I would rename them all to web ui

wesrisenmay-mx (approved):
Nice work

PR #34, opened 2024-08-13
Turborepo shared package

tphalp:
This introduces a shared-utils package as part of the monorepo. This will allow us to easily share code between apps.

Resolves UCP-325


apps/authentication-service/src/resources/clients/clientsRoutes.ts
```diff
@@ -17,22 +17,42 @@ export const clientsRoutes = (app: Application): void => {
 
   clientsRouterV1.post(
     "/keys",
-    [validateAccessToken, requiredScopes(WidgetHostPermissions.CREATE_KEYS)],
+    [
+      validateAccessToken,
+      requiredScopes(
+        (WidgetHostPermissions as Record<string, string>).CREATE_KEYS,
+      ),
```
tphalp, line 24: I was getting linting errors on this, unless I specified the type for the `WidgetHostPermissions` enum. Very strange, but I gave up trying to figure out why.

The same goes for the other uses of `WidgetHostPermissions` in this file.

apps/authentication-service/package.json
```diff
@@ -12,6 +12,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@repo/shared-utils": "^0.0.1",
```
wesrisenmay-mx, line 15: I've never seen a version here before: https://turbo.build/repo/docs/core-concepts/internal-packages

I've only seen "@repo/ui": "*"

apps/authentication-service/src/resources/clients/clientsRoutes.ts
wesrisenmay-mx: This sounds like there's something wrong with the way the package is setup with typescript

apps/ucp-ui/package.json
```diff
@@ -22,6 +22,7 @@
     "@mui/lab": "^5.0.0-alpha.173",
     "@mui/material": "^5.16.6",
     "@reduxjs/toolkit": "^2.2.7",
+    "@repo/shared-utils": "^0.0.1",
```
wesrisenmay-mx, line 25: This isn't actually used in ucp-ui right?

.github/workflows/auth-e2e-tests.yml
```diff
@@ -39,6 +39,9 @@ jobs:
 
           cat ${ENV_FILE_PATH}
 
+      - name: Build shared-utils
```
wesrisenmay-mx, line 42: Why do you have to do this?

apps/authentication-service/src/resources/clients/clientsRoutes.ts
wesrisenmay-mx: We should fix this. These shared enums aren't that useful if we don't know their types.

wesrisenmay-mx: Are you using the Just In time strategy? https://turbo.build/repo/docs/core-concepts/internal-packages#just-in-time-packages

tphalp: That strange thing is that in other places in the code where these are used, the typing isn't an issue. It was only in that one file, where I run into eslint complaining.

I'll look at the JIT stuff you linked to. I'm not sure if I'm using it.

apps/ucp-ui/package.json
tphalp: It is. Check `App.tsx`

tphalp: It is now.

.github/workflows/auth-e2e-tests.yml
tphalp: I may not have to. I tried getting turbo to manage it, through `dependsOn`, in the turbo.json file, but I couldn't get it to work as I expected it to work. I'm going to try again so we don't need this step in the workflow files.

apps/authentication-service/src/resources/clients/clientsRoutes.ts
tphalp: I just looked the JIT link. The one complication I can see, is that our node services aren't compiled, so they don't use webpack, turbopack, or vite, which is what that article suggests is necessary for JIT to work.

I'll do some more digging to see how I might be able to do this with node services.

.github/workflows/auth-e2e-client-tests.yml
```diff
@@ -43,7 +43,9 @@ jobs:
           project: apps/authentication-service
           start: npm run dev --workspace apps/authentication-service
           wait-on: "http://localhost:8089/ping"
-          spec: apps/authentication-service/cypress/e2e/**/clients.cy.ts
+          spec: |
+            apps/authentication-service/cypress/e2e/clients.cy.ts
+            apps/authentication-service/cypress/e2e/express.cy.ts
```
tphalp, line 48: Added `express.cy.ts`, as it was inadvertently removed in a previous PR.

package.json
```diff
@@ -16,17 +16,18 @@
     "lint": "turbo lint",
     "prettier": "turbo run prettier",
     "test": "turbo run test",
-    "prepare": "husky install"
+    "prepare": "husky"
```
tphalp, line 19: `husky install` is deprecated, and just `husky` is the current preferred command.

packages/shared-utils/src/consts.ts
```diff
@@ -0,0 +1,17 @@
...
+  READ_KEYS: "read:keys",
+  CREATE_KEYS: "create:keys",
+  DELETE_KEYS: "delete:keys",
+  ROTATE_KEYS: "rotate:keys",
+};
+
+export const WidgetHostPermissions = {
+  READ_WIDGET_ENDPOINTS: "read:widget-endpoints",
+  WRITE_WIDGET_ENDPOINTS: "write:widget-endpoints",
+};
```
tphalp, line 17: Enums didn't play well with the webpack loader. Converting these to consts fixed the issue when trying to use the shared package with the ucp-ui.

wesrisenmay-mx (approved):

PR #37, opened 2024-08-16
Add client grant to create flow

tphalp:
This PR adds the proper API Grant and Permissions (scope) for a newly-created Client.


apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -53,6 +69,8 @@ describe("Client API", () => {
   });
 
   it("clears the client_id from user metadata, tries creating a client without access token, creates a client, fails if another client request is made, gets the newly created client, and deletes the client", () => {
+    cy.wait(2000); // Because sometimes we hit the shortened rate limit
```
wesrisenmay-mx, line 72: auth0 is who is rate limiting us right?

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -1,14 +1,21 @@
...
   let newClientId: string;
   let newClientSecret: string;
+
   const keysUrl = `http://localhost:${PORT}/v1/clients/keys`;
+  const auth0BaseUrl = Cypress.env("AUTH0_DOMAIN") as string;
+  const auth0ClientGrantsUrl = `${auth0BaseUrl}/api/v2/client-grants`;
+  const clientScope = `${Object.values(WidgetHostPermissions).join(" ")}`;
 
   const getTokens = () => {
     cy.window()
```
wesrisenmay-mx, line 21: There's a lot of duplication in this function

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -69,16 +87,48 @@ describe("Client API", () => {
...
-    });
+        expect(response.status).to.eq(200);
+        expect(Object.keys(body)).to.have.length(2);
+        expect(Object.keys(body)).to.include("clientId");
+        expect(Object.keys(body)).to.include("clientSecret");
+
+        return newClientId;
+      })
+      .then((newClientId) => {
+        cy.request({
```
wesrisenmay-mx, line 104: You're not returning this request, so how is that last then on 114 getting a response?

apps/authentication-service/cypress/e2e/clients.cy.ts
tphalp: No, it's a self-imposed rate-limit for testing. See the `rateLimiting.cy.ts` file.

tphalp: There's no way to change the rate limit on the fly, so it affects all cypress tests.

wesrisenmay-mx: I've seen the rate limiting tests, but where is the rate limit enforced?

apps/authentication-service/cypress/e2e/clients.cy.ts
wesrisenmay-mx: Ah good call. Can't see clearly enough without having an ide

apps/authentication-service/cypress/e2e/clients.cy.ts
tphalp: It's in the initialization of the service. It's an express middleware.

tphalp: The rate is much lower if the ENV === "test"

apps/authentication-service/cypress/e2e/clients.cy.ts
tphalp: I cleaned that up a bunch. Let me know what you think.

apps/authentication-service/cypress/e2e/clients.cy.ts
tphalp: For anyone reading this after the fact, we're going to test some options of splitting the rate-limiting stuff into its own isolated context (separate workflow, etc), so the low, test-specific rate limits won't affect other tests.

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -220,4 +258,38 @@ describe("Client API", () => {
...
+      headers: {
+        ContentType: "application/json",
+        Authorization: `Bearer ${accessToken}`,
+      },
+    }).then(() => {
+      cy.request({
+        url: `http://localhost:8088/institutions/cacheList`,
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${accessTokenWidget}`,
```
wesrisenmay-mx, line 275: This isn't testing that the application you just created has access to what it's supposed to. This is using a token from something that is kind of fake. This doesn't give me confidence that the create keys endpoint creates an application with the access that it needs.

tphalp: Okay, new code that actually tests with the newly created client.

apps/authentication-service/cypress.example.env.json
```diff
@@ -4,5 +4,8 @@
   "AUTH_USERNAME_WITH_KEY_ROLES": "tyson.phalp+cypress@mx.com",
   "AUTH_PASSWORD_WITH_KEY_ROLES": "",
   "AUTH_USERNAME_WITHOUT_KEY_ROLES": "tyson.phalp+cypressbasic@mx.com",
-  "AUTH_PASSWORD_WITHOUT_KEY_ROLES": ""
+  "AUTH_PASSWORD_WITHOUT_KEY_ROLES": "",
+  "WIDGET_CLIENT_ID": "U1S5r5EQ9bPqXD5ai6769u9oi63BD1S2",
```
wesrisenmay-mx, line 8: We don't need this anymore right?

tphalp: Correct. Removed...

wesrisenmay-mx (approved):

PR #38, opened 2024-08-16
remove supports_account_statement, add supports_aggregation

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-303

Issue: https://universalconnect.atlassian.net/browse/UCP-302


wesrisenmay-mx:
Can we make our e2e tests for the institution service better by checking that the correct properties are showing up in the response?

wesrisenmay-mx (approved):

PR #39, opened 2024-08-19
Wes/view api keys

wesrisenmay-mx:


apps/ucp-ui/cypress.config.ts
```diff
@@ -3,6 +3,7 @@ const { defineConfig } = require("cypress");
 module.exports = defineConfig({
   e2e: {
     baseUrl: "http://localhost:3000/",
+    retries: 1,
```
tphalp, line 6: Is this because of the rate limit issue?

wesrisenmay-mx: Kind of, but having a retry is generally a healthy practice.

PR #41, opened 2024-08-19
Isolate Rate-limit tests

tphalp:
This PR has a few other fixes. While working on this PR, I realized that the Auth E2E tests were broken. The workflow had a syntax error. The syntax error caused an issue where the auth workflow didn't even show up in the required checks for any PRs. The syntax error was introduced in my [last PR](https://github.com/Universal-Connect-Project/ucp-hosted-app/pull/37). 

Because of this error, I thought the auth E2E tests were fine, but they weren't . I had to add some additional code to get the institution service working for some of the auth tests. Those changes are part of this PR.

**Additional context to this PR**: We have been running into rate-limits from Auth0. You can see Auth0's rate limits [here](https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/rate-limit-configurations/free-public). 

I originally found [this link](https://docs.cypress.io/guides/end-to-end-testing/auth0-authentication#Auth0-Rate-Limiting-Logins) on the Cypress docs about how to deal with the Auth0 Rate Limits. 

However, after implementing this "fix", it still didn't finish consistently. I then found [this GitHub issue](https://github.com/cypress-io/cypress-documentation/issues/4103) on the cypress project. It is an issue created by someone at Auth0, stating that the suggestion on Cypress' docs about circumventing the Auth0 Rate Limits is inaccurate, and will not help.

After all of this, the only thing that gave any increase is success, is adding `cy.wait(x)` to several places in the e2e test file. After doing this, I had a roughly 85% - 90% success rate with successive workflow runs. Previously, it was roughly a 35% - 40% success rate, on successive runs. 

When we start paying for Auth0, we will also get[ improved rate limits](https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/rate-limit-configurations/self-service-public), and this won't be as big of a problem.


.github/workflows/auth-e2e-tests.yml
```diff
@@ -23,43 +37,44 @@ jobs:
...
+          echo WIDGET_CLIENT_SECRET=${{ secrets.WIDGET_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
+          echo WEB_UI_CLIENT_SECRET=${{ secrets.WEB_UI_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
+          echo E2E_INSTITUTION_PASSWORD=${{ secrets.E2E_INSTITUTION_PASSWORD }} >> ${ENV_FILE_PATH}
+
+          # Copy the env file to the institution service
+          cp ${ENV_FILE_PATH} ${INST_ENV_FILE_PATH} 
 
       - name: Start Institution Service
-        run: npm run dev --workspace apps/institution-service
+        run: npm run dev --workspace apps/institution-service &
```
tphalp, line 69: Without the `&`, the tests do not continue. GH thinks the command hasn't finished, so the workflow gets stuck, and never finishes.

apps/authentication-service/cypress.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
     AUTH0_CLIENT_AUDIENCE: process.env.AUTH0_CLIENT_AUDIENCE,
     AUTH0_CLIENT_ID: process.env.AUTH0_CLIENT_ID,
     AUTH0_CLIENT_SECRET: process.env.AUTH0_CLIENT_SECRET,
+    AUTH0_WIDGET_AUDIENCE: process.env.AUTH0_WIDGET_AUDIENCE,
```
tphalp, line 12: This is needed for the institution service to be able to start successfully.

apps/authentication-service/cypress/e2e/express.cy.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  it("handles not found", () => {
+    cy.request({
+      url: `http://localhost:${PORT}/pig`,
+      failOnStatusCode: false,
+    })
+      .its("status")
+      .should("eq", 404);
+  });
+});
```
tphalp, line 21: New file that includes the basic expressjs tests. These were pulled out of the rateLimit e2e test file.

apps/authentication-service/src/init.ts
```diff
@@ -4,12 +4,12 @@ import nocache from "nocache";
 import { rateLimit } from "express-rate-limit";
 import express, { Application, NextFunction, Request, Response } from "express";
 
-import envs from "./config";
 import { errorHandler } from "@/middleware/errorMiddleware";
 import { notFoundHandler } from "@/middleware/notFoundMiddleware";
 import { clientsRoutes } from "@/resources/clients/clientsRoutes";
 
 const rateLimitWindowMinutes = 10;
+const rateLimitTest = process.env.RATE_LIMIT_TEST === "true" || false;
```
tphalp, line 12: This env var is only passed into the code in GH Actions.

apps/authentication-service/.env.example
```diff
@@ -8,6 +8,7 @@ CLIENT_ORIGIN_URL=http://localhost:3031
 AUTH0_DOMAIN=dev-d23wau8o0uc5hw8n.us.auth0.com
 AUTH0_M2M_AUDIENCE=https://dev-d23wau8o0uc5hw8n.us.auth0.com/api/v2/
 AUTH0_CLIENT_AUDIENCE=ucp-hosted-apps
+AUTH0_WIDGET_AUDIENCE=ucp-widget-interactions
```
tphalp, line 11: Fixes local e2e test for auth/inst

.github/workflows/auth-e2e-tests.yml
```diff
@@ -23,43 +37,44 @@ jobs:
...
-          spec: apps/authentication-service/cypress/e2e/**/!(rateLimiting.cy.ts)
-
-      - name: Kill server # In order to reset rate-limiting
-        run: kill $(lsof -t -i:8089)
-
-      - name: Cypress run (Rate-limiting)
         uses: cypress-io/github-action@v6
         with:
           project: apps/authentication-service
           start: npm run dev --workspace apps/authentication-service
```
wesrisenmay-mx, line 75: You should start the institution service here

.github/workflows/auth-rate-limit-e2e-tests.yml
```diff
@@ -0,0 +1,57 @@
+name: Auth E2E (Rate-limit)
```
wesrisenmay-mx, line 1: Based on what I'm seeing it looks like if you were to run `npm run cypress:open --workspace=apps/authentication-service`, then it would show the rate limiting e2e tests as well as the express e2e tests. IMO we should set it up like it's set up for ucw-app. You have to run separate commands to open up the different suites. We don't want someone to think there's a failure because they don't have the correct configuration for the test to pass.

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -41,17 +40,13 @@ describe("Client API", () => {
...
-    getLocalStorage({
-      storageKey: "jwt-auth-m2m",
-      callback: (token: string) => {
-        accessTokenM2M = token;
-      },
-    });
   };
 
   before(() => {
+    cy.wait(61000);
```
tphalp, line 46: This is painful, but anything less than this resulted in substantially less successful runs.

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -41,17 +40,13 @@ describe("Client API", () => {
...
+
     getTokens();
-    if (!accessToken || !accessTokenBasic || !accessTokenM2M) {
+    if (!accessToken || !accessTokenBasic) {
```
tphalp, line 49: Realized `accessTokenM2M` was not being used in this file anymore.

apps/authentication-service/cypress/e2e/clients.cy.ts
```diff
@@ -247,6 +252,8 @@ describe("Client API", () => {
   });
 
   it("creates a new client, accesses institution cache endpoint, and deletes the client", () => {
+    cy.wait(10000);
```
tphalp, line 255: Without this wait, this series of request would fail the majority of the time, with a rate limit error.

apps/authentication-service/package.json
```diff
@@ -4,8 +4,11 @@
   "description": "UCP Authentication Service",
   "scripts": {
     "dev": "npx -y nodemon -r tsconfig-paths/register ./src/index.ts",
-    "cypress:open": "cypress open",
-    "cypress:run": "cypress run",
+    "dev:rate-limit": "RATE_LIMIT_TEST=true npm run dev",
+    "cypress:open": "cypress open --config specPattern='cypress/e2e/**/!(rateLimiting.cy.ts)'",
+    "cypress:run": "cypress run --config specPattern='cypress/e2e/**/!(rateLimiting.cy.ts)'",
+    "cypress:open:rate-limit": "cypress open --config specPattern='cypress/e2e/rateLimiting.cy.ts'",
+    "cypress:run:rate-limit": "cypress run --config specPattern='cypress/e2e/rateLimiting.cy.ts'",
```
tphalp, line 11: New scripts for isolation of our own express rate-limit tests.

.github/workflows/auth-e2e-tests.yml
```diff
@@ -23,44 +37,47 @@ jobs:
...
+          echo WIDGET_CLIENT_SECRET=${{ secrets.WIDGET_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
+          echo WEB_UI_CLIENT_SECRET=${{ secrets.WEB_UI_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
+          echo E2E_INSTITUTION_PASSWORD=${{ secrets.E2E_INSTITUTION_PASSWORD }} >> ${ENV_FILE_PATH}
+
+          # Copy the env file to the institution service
+          cp ${ENV_FILE_PATH} ${INST_ENV_FILE_PATH} 
 
-      - name: Start Institution Service
-        run: npm run dev --workspace apps/institution-service
+#      - name: Start Institution Service
```
wesrisenmay-mx, line 68: Delete

apps/authentication-service/cypress/e2e/clients.cy.ts
wesrisenmay-mx: Waiting 60 seconds isn't reasonable. Are we being rate limited by accounts, or by ips, or both?

tphalp: See my PR description for more info. I believe we are limited by IP.

wesrisenmay-mx (approved):

PR #42, opened 2024-08-26
Remove hard-coded clientid/secret in institution service e2e tests

tphalp:
This PR changes the e2e tests for the Institution Service to use the user-flow, instead of using a hard-coded ClientID and ClientSecret.


apps/authentication-service/cypress/support/commands.ts
```diff
@@ -75,7 +75,6 @@ const login = (args: LoginArgs) => {
           scope,
           audience,
           client_id: clientId,
-          client_secret: clientSecret,
```
tphalp, line 78: I realized this wasn't needed, but ID is.

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -62,7 +130,7 @@ describe("Institution endpoints", () => {
         failOnStatusCode: false,
         method: "GET",
         headers: {
-          Authorization: `Bearer ${Cypress.env("USER_ACCESS_TOKEN")}`,
+          Authorization: `Bearer ${Cypress.env("NO_KEYS_USER_ACCESS_TOKEN")}`,
```
wesrisenmay-mx, line 133: Where is this coming from? Won't this token expire?

apps/institution-service/cypress/support/commands.ts
```diff
@@ -1,46 +1,86 @@
...
+  grantType: "password" | "client_credentials";
+  audience: string;
+  clientId: string;
+  clientSecret: string;
+  scope?: string;
+  usernameEnvKey?: string;
+  passwordEnvKey?: string;
+};
+
+const login = (args: LoginArgs) => {
```
wesrisenmay-mx, line 18: Am I crazy or is half this function unused?

apps/institution-service/cypress/e2e/institutions.cy.ts
tphalp: It comes from this code: https://github.com/Universal-Connect-Project/ucp-hosted-app/pull/42/files/7235853217b8e9ac5a7a30f97027d350c456a456#diff-b52e65dd5f7d947e1df05c99feab0f6bc6a36f97d6f02312e3e706c99ea107cfR67-R79

Not hard-coded. This is code that Craig originally wrote, I just renamed the variable to be more specific about what it is.

wesrisenmay-mx: Ah I see now. We're good.

apps/institution-service/cypress/support/commands.ts
tphalp: I pulled this over from the auth service, which uses all of it. I'll pare it down since the institution service doesn't do any other grant-types.

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -1,48 +1,179 @@
...
 import { CachedInstitution } from "../../src/tasks/loadInstitutionsFromJson";
 
+const keysUrl = `http://localhost:8089/v1/clients/keys`;
+
+type Keys = {
+  clientId: string;
+  clientSecret: string;
+};
+
+const getLocalStorage = (args: {
```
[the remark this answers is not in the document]
wesrisenmay-mx: The whole login process seems more complex than it needs to be. There is a getLocalStorage function, we're setting something in local storage via a command, and the login function is more complex than it needs to be.

We would have a lot less lines of code if we exported a paired down version of the login function and grabbed the access token directly instead of storing and retrieving it from localstorage

apps/institution-service/cypress/support/commands.ts
tphalp: I refactored it to be simpler.

tphalp: I just saw your other comment. I'll move it into the main spec file so we don't have to mess with this.

apps/institution-service/cypress/support/e2e.ts
```diff
@@ -16,43 +17,65 @@
...
+      client_id: ucpWebUiClientId,
       username,
       password,
-      audience: audience,
-      client_id: ucpWebUiClientId,
-      client_secret: ucpWebUiClientSecret,
     },
   }).then((response: Cypress.Response<JwtPayload>) => {
-    Cypress.env("USER_ACCESS_TOKEN", response.body.access_token);
+    Cypress.env("NO_KEYS_USER_ACCESS_TOKEN", response.body.access_token);
```
wesrisenmay-mx, line 79: The NO_KEYS name is confusing. It makes it seem like this is a user that doesn't have api key access. I believe what this test is supposed to be testing is that if a user requests with all the right information to get the cacheList that they can't get it. We should request with all the scope and audience necessary to try and get access when we make this request.

tphalp: This code is basically untouched from what existed before this PR. I just renamed some things that I thought were too vague. 

The token that comes from this support function shouldn't have the rights to hit the cacheList endpoint. The test that uses this token is testing that a 403 should be returned.

I'll rename this token, since it actually represents a user who does not have access to interact with the widget api (in the auth0 context)

wesrisenmay-mx (approved):

PR #44, opened 2024-09-06
Wes/navigation

wesrisenmay-mx:


apps/ucp-ui/src/Routes.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const mockLogout = jest.fn();
+
+jest.mock("@auth0/auth0-react", () => ({
+  useAuth0: () => ({
+    logout: mockLogout,
+  }),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  withAuthenticationRequired: (component: any): any => component,
+}));
```
Craiting, line 14: Is this used indirectly in the test? Why do you need to mock it?

wesrisenmay-mx: Rendering the children requires these. I improved it to use an __mock__

Craiting (approved):

PR #45, opened 2024-09-10
Wes/institutions

wesrisenmay-mx:


apps/ucp-ui/src/Layout/SideNav.tsx
```diff
@@ -28,6 +36,11 @@ const SideNav = () => {
   const { pathname } = useLocation();
 
   const links = [
+    {
+      label: SIDE_NAV_INSTITUTIONS_LINK_TEXT,
+      Icon: AccountBalanceOutlined,
```
tphalp, line 41: Does `Icon` need to be a different case from the other properties here?

tphalp (commented):
One comment

apps/ucp-ui/src/Layout/SideNav.tsx
wesrisenmay-mx: Yeah it does. React won't let you render anything unless it starts with a capital letter

tphalp (approved):

PR #51, opened 2024-09-13
Switch to using app_metadata for api keys

tphalp:


wesrisenmay-mx (approved):
You're planning on manually cleaning up all the accounts that already have client ids right?

tphalp:
> You're planning on manually cleaning up all the accounts that already have client ids right?

I already did

PR #53, opened 2024-09-13
Wes/institution permissions endpoint

wesrisenmay-mx:


.github/workflows/institution-e2e-tests.yml
```diff
@@ -58,6 +58,8 @@ jobs:
           echo WEB_UI_CLIENT_SECRET=${{ secrets.WEB_UI_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
           echo E2E_INSTITUTION_PASSWORD=${{ secrets.E2E_INSTITUTION_PASSWORD }} >> ${ENV_FILE_PATH}
           echo SUPER_ADMIN_PASSWORD=${{ secrets.SUPER_ADMIN_PASSWORD }} >> ${ENV_FILE_PATH}
+          echo AGGREGATOR_ADMIN_USERNAME=${{ secrets.AGGREGATOR_ADMIN_USERNAME }} >> ${ENV_FILE_PATH}
```
Craiting, line 61: Usually we put the non-sensitive stuff into a variable then you don't need to do this.

apps/institution-service/src/controllers/permissionController.ts
```diff
@@ -0,0 +1,24 @@
...
+  const decodedToken = jwtDecode<DecodedToken>(token as string);
+
+  const permissions = {
+    canCreateInstitution: !!decodedToken?.permissions?.find((permission) =>
+      [
+        UiUserPermissions.CREATE_INSTITUTION,
+        UiUserPermissions.CREATE_INSTITUTION_AGGREGATOR,
+      ].includes(permission),
+    ),
+  };
```
Craiting, line 20: `some` returns true if any of the tests pass and I feel like it makes the code a little easier to read, what do you think of this change?

```suggestion
  const permissions = {
      canCreateInstitution: [
        UiUserPermissions.CREATE_INSTITUTION,
        UiUserPermissions.CREATE_INSTITUTION_AGGREGATOR,
      ].some((permission) => decodedToken?.permissions?.includes(permission)),
    };
```

wesrisenmay-mx: 👍

apps/institution-service/src/app.ts
```diff
@@ -14,6 +15,7 @@ app.use(logger("dev"));
 
 // Routes
 app.use("/institutions", institutionRoutes);
+app.use(permissionsRoutes);
```
Craiting, line 18: I think it's kind of nice to put the base route here just because then in the routes file you don't have to include `/permissions` in each route you define. You only have one route in permissionsRoutes so it's not repetitive in this case.

Craiting: It also limits the scope of permissionRoutes file. Otherwise you could technically add institution endpoints or anything else in there too.

wesrisenmay-mx: Done

Craiting (approved):

PR #55, opened 2024-09-18
Ct/db refactor

Craiting:
Database architecture change and renaming "provider" to "aggregator" everywhere.
image


apps/institution-service/src/models/aggregator.ts
```diff
@@ -0,0 +1,80 @@
...
+
+AggregatorIntegration.belongsTo(Institution, {
+  foreignKey: "institution_id",
+  as: "institution",
+});
+
+AggregatorIntegration.belongsTo(Aggregator, {
+  foreignKey: "aggregatorId",
+  as: "aggregator",
+});
```
tphalp, line 80: Should the instances of "institution" be "provider" now?

apps/institution-service/src/models/aggregatorIntegration.test.ts
```diff
@@ -0,0 +1,28 @@
+import { AggregatorIntegration } from "./aggregatorIntegration";
+
+describe("AggregatorIntegration Model", () => {
+  it("should create an aggregator", async () => {
+    const aggregatorAttributes = {
+      name: "mx",
+      supports_oauth: true,
+      supports_history: true,
+      institution_id: "123",
+      aggregator_institution_id: "mx_oauth_bank",
```
tphalp, line 10: Should "aggregator_institution_id" be "aggregator_provider_id"?

apps/institution-service/src/models/aggregator.ts
tphalp: I think I'm misunderstanding here....we're moving away from "provider" as meaning "aggregator", correct?

tphalp (approved):

apps/institution-service/migrations/20240918164405-renameAggregatorFields.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.renameColumn('aggregatorIntegrations', 'provider_institution_id', 'aggregator_institution_id')
```
wesrisenmay-mx, line 6: camel case

apps/institution-service/src/models/aggregator.ts
```diff
@@ -0,0 +1,80 @@
...
+import sequelize from "../database";
+import { AggregatorIntegration } from "./aggregatorIntegration";
+import { Institution } from "./institution";
+
+export class Aggregator extends Model<
+  InferAttributes<Aggregator, { omit: "aggregatorIntegrations" }>,
+  InferCreationAttributes<Aggregator, { omit: "aggregatorIntegrations" }>
+> {
+  declare id?: number;
+  declare name: string;
```
wesrisenmay-mx, line 20: Isn't this required to be unique in the widget? It seems like this should be the primary key

PR #56, opened 2024-09-20
Change ucp_id to id and use UUID

Craiting:


apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
```diff
@@ -96,14 +101,29 @@ async function loadInstitutionData() {
...
+        institution_id: institution.id,
         supports_oauth: finicity.supports_oauth,
         supports_identification: finicity.supports_identification,
         supports_verification: finicity.supports_verification,
         supports_history: finicity.supports_history,
         supports_aggregation: finicity.supports_aggregation,
       } as AggregatorIntegration);
     }
+
+    if (institution?.testExampleA?.id) {
```
wesrisenmay-mx, line 113: Don't we need testExampleB?

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
```diff
@@ -96,14 +101,29 @@ async function loadInstitutionData() {
...
+      const testExampleA = institution.testExampleA;
+      aggregatorList.push({
+        isActive: false,
```
wesrisenmay-mx, line 116: I think we want the actual isActive value here

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
Craiting: oh yeah is that the only other one

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
Craiting: Good catch

Craiting: I just realized the json file doesn't include the isActive field, should we add that?

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
Craiting: I think I've added all the necessary test institutions to the json file now

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
Craiting: I don't think ucw-app uses it at all so it's probably not necessary to include in the cacheList endpoint right?

wesrisenmay-mx: No we don't need to return it in the cacheList endpoint, because we don't return institutions that aren't active in the cacheList.

I put isActive to false on Finicity because they aren't active, and so that I could use that for tests. I think ideally the institution map file has isActive, so we can run tests from a solid base and test institutions that aren't active.

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
wesrisenmay-mx: Yeah testExampleB is the only other one.

PR #61, opened 2024-09-24
Ct/institution pagination

Craiting:


apps/institution-service/openApi.json
```diff
@@ -21,6 +21,195 @@
     }
   ],
   "paths": {
+    "/institutions": {
```
wesrisenmay-mx, line 24: There's already an institutions path below for "post"

apps/institution-service/src/controllers/institutionController.ts
```diff
@@ -97,3 +98,84 @@ export const updateInstitution = async (req: Request, res: Response) => {
...
+  currentPage: number;
+  pageSize: number;
+  totalRecords: number;
+  totalPages: number;
+  institutions: InstitutionDetail[];
+}
+
+export const getPaginatedInstitutions = async (req: Request, res: Response) => {
+  try {
+    const { limit, offset, page } = res.locals.pagination! as PaginationOptions;
```
wesrisenmay-mx, line 140: I feel like not using middleware would be simpler and better.

Looking at this code I had to look up what the heck locals are, and go figure out how we were attaching locals to the response.

I'd prefer something like const { limit, offset, page } = getPaginationOptions(req)

wesrisenmay-mx: It also makes testing more realistic, because our tests will use the actual request object instead of the assuming the locals object is populated correctly

apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
```diff
@@ -43,21 +43,21 @@ async function loadInstitutionData() {
   const institutionsList: Institution[] = [];
   const aggregatorList: AggregatorIntegration[] = [];
   const [mxAggregator, _mxCreated] = await Aggregator.findOrCreate({
-    where: { name: "mx" },
+    where: { name: "mx", displayName: "MX" },
```
wesrisenmay-mx, line 46: We should add the logos

Craiting: What are the urls for the logos?

apps/institution-service/src/controllers/institutionController.ts
Craiting: That's how I had it originally until I moved it into a middleware, I'll switch it back

apps/institution-service/src/controllers/institutionController.ts
```diff
@@ -97,3 +98,98 @@ export const updateInstitution = async (req: Request, res: Response) => {
...
+  institutions: InstitutionDetail[];
+}
+
+interface PaginationOptions {
+  page: number;
+  limit: number;
+  offset: number;
+}
+
+const getPaginationOptions = (req: Request): PaginationOptions => {
```
wesrisenmay-mx, line 144: I'd write tests for this function as well. At least to check the default behavior

wesrisenmay-mx (approved):

PR #64, opened 2024-09-30
Wes/institutions use api

wesrisenmay-mx:


apps/ucp-ui/src/Institutions/Institutions.tsx
```diff
@@ -1,40 +1,137 @@
...
+            [prop]: true,
+          }),
+          {},
+        ),
+        aggregator: {
+          displayName: (Math.random() + 1).toString(36).substring(7),
+        },
+      })),
+  }));
+};
```
tphalp, line 64: Is this method so the skeletons work prior to data being available?

tphalp (dismissed):

apps/ucp-ui/src/Institutions/Institutions.tsx
wesrisenmay-mx: Yeah i'm just adding some variance so it looks somewhat real

tphalp (approved):

PR #66, opened 2024-10-01
Add edit aggregatorIntegration endpoint

Craiting:


apps/institution-service/src/middlewares/validationMiddleware.ts
```diff
@@ -86,3 +97,47 @@ export const validateUserCanEditInstitution = async (
     return res.status(500).json({ error: "Error validating user permission" });
   }
 };
+
+export const validateUserCanEditAggregatorIntegration = async (
```
wesrisenmay-mx, line 101: There's a lot of logic in here. Need tests

apps/institution-service/src/app.ts
```diff
@@ -30,6 +33,16 @@ const cacheListLimiter = createLimiter({
   requestLimit: 3,
 });
 
+sequelize
+  .sync()
+  .then(() => {
+    defineAssociations();
```
Craiting, line 39: I ran into a really weird problem where certain defined associations were working while testing manually, but failing in the specs. Did some research and it seems like it's best practice to load the associations after all the models are initialized and before running tests or the server.

wesrisenmay-mx (approved):

PR #68, opened 2024-10-03
Add create aggregatorIntegration endpoint

Craiting:


apps/institution-service/src/controllers/aggregatorIntegrationController.ts
```diff
@@ -38,3 +42,30 @@ export const updateAggregatorIntegration = async (
...
+  } catch (error) {
+    if (error instanceof ValidationError) {
+      res.status(400).json({
+        error: "Invalid aggregatorIntegration Data",
+        message: error.errors[0]?.message,
+      });
+    } else if (error instanceof Error) {
+      res.status(400).json({ error: "Database Error", message: error.message });
+    } else {
+      res.status(500).json({ error: "Error", message: "something went wrong" });
```
tphalp, line 68: This is a nitpick, and not that important, but I'd use sentence case in any error message, just for consistency. "Something went wrong"

tphalp (approved):

apps/institution-service/cypress/e2e/aggregatorIntegrations.cy.ts
```diff
@@ -223,3 +229,124 @@ describe("PUT /aggregatorIntegrations/:id (AggregatorIntegration update)", () =>
     method: "PUT",
   });
 });
+
+describe.only("POST /aggregatorIntegrations (Create)", () => {
```
wesrisenmay-mx, line 233: only?

apps/institution-service/openApi.json
```diff
@@ -761,6 +761,118 @@
             }
           }
         }
+      },
+      "post": {
```
wesrisenmay-mx, line 765: It looks like this is attached to the wrong url. When I load it in the swagger editor it says it requires an id

image

apps/institution-service/src/controllers/aggregatorIntegrationController.ts
wesrisenmay-mx: Agreed

apps/institution-service/src/controllers/aggregatorIntegrationController.ts
```diff
@@ -38,3 +42,30 @@ export const updateAggregatorIntegration = async (
     });
   }
 };
+
+export const createAggregatorIntegration = async (
```
wesrisenmay-mx, line 46: This is missing tests

wesrisenmay-mx (approved):

PR #70, opened 2024-10-08
Wes/institution permissions

wesrisenmay-mx:


apps/institution-service/src/shared/utils/permissionValidation.test.ts
```diff
@@ -231,4 +233,105 @@ describe("permissionValidation", () => {
       ).toBe(EditAggregatorIntegrationValidationErrorReason.NotYourAggregator);
     });
   });
+
+  describe("validateUserCanCreateAggregatorIntegration", () => {
+    it("returns true if they are a super admin and they there are aggregators without an integration", async () => {
```
Craiting, line 238:
```suggestion
    it("returns true if they are a super admin and there are aggregators without an integration", async () => {
```

apps/institution-service/src/shared/utils/permissionValidation.test.ts
```diff
@@ -231,4 +233,105 @@ describe("permissionValidation", () => {
...
+              authorization: createTestAuthorization({
+                permissions: [UiUserPermissions.CREATE_AGGREGATOR_INTEGRATION],
+              }),
+            },
+          } as Request,
+        }),
+      ).toBe(true);
+    });
+
+    it("returns false if they are a super admin and they there aren't aggregators without an integration", async () => {
```
Craiting, line 253: Same typo with "they there"

apps/institution-service/src/shared/utils/permissionValidation.test.ts
wesrisenmay-mx: Fixed

apps/institution-service/src/shared/utils/permissionValidation.test.ts
wesrisenmay-mx: Fixed

Craiting (approved):

PR #72, opened 2024-10-08
Restrict create aggregator integration from having multiple of the same agg/ins combo

Craiting:
Doesn't allow you to create multiple aggregatorIntegrations for the same aggregator/institution combo

### DIFF
--- a/apps/institution-service/cypress/e2e/aggregatorIntegrations.cy.ts
+++ b/apps/institution-service/cypress/e2e/aggregatorIntegrations.cy.ts
@@ -1,3 +1,4 @@
+import { UUID } from "crypto";
 import { PORT } from "shared/const";
 import {
   testExampleAAggregatorId,
@@ -10,6 +11,7 @@ import {
 } from "../shared/constants/accessTokens";
 import { createAuthorizationHeader } from "../shared/utils/authorization";
 import {
+  createTestInstitution,
   deleteAggregatorIntegration,
   runInvalidPermissionCheck,
   runTokenInvalidCheck,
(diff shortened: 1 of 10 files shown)


apps/institution-service/src/middlewares/validationMiddleware.ts
```diff
@@ -203,6 +206,28 @@ export const validateUserCanCreateAggregatorIntegration = async (
...
   const aggregatorRequestBody = req.body as AggregatorIntegrationRequestBody;
 
+  if (aggregator.id !== aggregatorRequestBody.aggregatorId) {
+    return res.status(403).json({
+      error:
+        "An Aggregator cannot create an aggregatorIntegration belonging to another aggregator",
+    });
+  }
+
+  const aggregatorIntegrationForThisInstitutionAndAggregator =
```
wesrisenmay-mx, line 216: We don't need this anymore right?

apps/institution-service/cypress/e2e/aggregatorIntegrations.cy.ts
```diff
@@ -363,43 +418,54 @@ describe("POST /aggregatorIntegrations (Create)", () => {
       ) => {
         expect(response.status).to.eq(400);
 
-        expect(response.body.error).to.eq("Database Error");
+        expect(response.body.error).to.eq(
+          "Invalid reference in the field: providers_institution_id_fkey.",
```
Craiting, line 422: "provider" is left over from a database index that was created before the name change. I think when I fix the camelCase fields I will wipe everything and rebuild the schema from scratch and this will be renamed to "aggregator"

apps/institution-service/src/controllers/aggregatorIntegrationController.ts
```diff
@@ -15,6 +15,30 @@ interface AggregatorIntegrationParams {
   isActive: boolean;
 }
 
+function getErrorCodeAndMessage(error: Error): [number, string] {
+  if (error instanceof ValidationError) {
+    if (error.name === "SequelizeUniqueConstraintError") {
+      return [
+        409,
+        "An AggregatorIntegration for that Institution/Aggregator already exists. Cannot duplicate",
```
wesrisenmay-mx, line 23: I'd remove the cannot duplicate. It makes it seem like this is a duplicate request

wesrisenmay-mx (approved):

PR #74, opened 2024-10-10
Add aggregator list endpoint

Craiting:


apps/institution-service/cypress/e2e/aggregators.cy.ts
```diff
@@ -0,0 +1,86 @@
+import { Aggregator } from "models/aggregator";
```
tphalp, line 1: Add test(s) for potential error paths.

wesrisenmay-mx: Is there an error path other than an invalid token?

apps/institution-service/cypress/e2e/aggregators.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+        (attribute) => {
+          expect(aggregator).to.haveOwnProperty(attribute);
+        },
+      );
+    });
+  });
+};
+
+describe("/aggregators GET", () => {
+  it("returns a list of aggregators for a super admin", () => {
```
wesrisenmay-mx, line 33: Because there is no scope checking I'd just do 2 tests. 1 for valid token and 1 for invalid token. We don't have any logic that separates the roles on this endpoint.

apps/institution-service/src/controllers/aggregatorController.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        authorization: createTestAuthorization({
+          permissions: [],
+        }),
+      },
+    } as Request;
+    let responseData: { aggregators: object[] } = { aggregators: [] };
+    const res = {
+      json: jest
+        .fn()
+        .mockImplementation(
```
wesrisenmay-mx, line 22: Are you sure you need to do this mock implementation?

Craiting: It's probably not necessary

wesrisenmay-mx (approved):

PR #75, opened 2024-10-14
remove logo required

Craiting:
Logo not required on Institution create. If one isn't provided then the front-end will use a default logo

ucw-app update: https://github.com/Universal-Connect-Project/ucw-app/pull/124

### DIFF
--- a/apps/institution-service/src/controllers/institutionController.test.ts
+++ b/apps/institution-service/src/controllers/institutionController.test.ts
@@ -378,9 +378,9 @@ describe("institutionController", () => {
                 supports_history: expect.any(Boolean),
                 isActive: expect.any(Boolean),
                 aggregator: expect.objectContaining({
-                  name: "mx",
+                  name: expect.any(String),
                   id: expect.any(Number),
-                  displayName: "MX",
+                  displayName: expect.any(String),
                   logo: "https://logo.com",
                 }),
               }),
--- a/apps/institution-service/src/controllers/institutionController.test.ts
+++ b/apps/institution-service/src/controllers/institutionController.test.ts
@@ -11,6 +11,7 @@ import {
   seedInstitutionName,
   testInstitution,
 } from "../test/testData/institutions";
+import { createTestAuthorization } from "../test/utils";
 import {
   createInstitution,
   getInstitution,
@@ -18,7 +19,6 @@ import {
   getPaginatedInstitutions,
   updateInstitution,
 } from "./institutionController";
-import { createTestAuthorization } from "../test/utils";
 
 const createNewInstitution = async () => {
   return await Institution.create(testInstitution);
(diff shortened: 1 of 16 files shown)


apps/institution-service/src/controllers/institutionController.test.ts
```diff
@@ -87,7 +87,9 @@ describe("institutionController", () => {
       // eslint-disable-next-line @typescript-eslint/unbound-method
       expect(res.status).toHaveBeenCalledWith(201);
       expect(res.json).toHaveBeenCalledWith(
-        expect.objectContaining(institutionBody),
+        expect.objectContaining({
+          institution: expect.objectContaining(institutionBody),
+        }),
```
tphalp, line 92: Just a question because of my ignorance. Is this a common pattern?--nested "expects"? Very cool. I didn't realize you could do that.

apps/ucp-ui/src/Institutions/ChangeInstitution/api.ts
```diff
@@ -15,7 +15,9 @@ export interface EditInstitutionParams extends CreateInstitution {
 }
 
 export interface Institution {
-  id: string;
+  institution: {
+    id: string;
+  };
```
tphalp, line 20: Why add the additional "institution" prop?

tphalp (commented):
Just a question

apps/ucp-ui/src/Institutions/utils.ts
```diff
@@ -4,3 +4,7 @@ export const aggregatorIntegrationsSortByName = (
   a: AggregatorIntegration,
   b: AggregatorIntegration,
 ) => a.aggregator.displayName.localeCompare(b.aggregator.displayName);
+
+export const isEmptyStr = (str: string | undefined) => {
```
wesrisenmay-mx, line 8: Why don't we store null instead of an empty string?

Craiting: I couldn't figure out a very clean way to do it. Let me try something different and let me know if you think it's better.

apps/institution-service/src/controllers/institutionController.test.ts
Craiting: I'm not sure how common it is but I've found it to be pretty useful

apps/ucp-ui/src/Institutions/ChangeInstitution/api.ts
Craiting: The institution create endpoint was updated to include the nested institution in the response to match the other endpoints. So this update is necessary to accommodate that update.

apps/ucp-ui/src/Institutions/ChangeInstitution/api.ts
```diff
@@ -39,7 +40,7 @@ const transformBody = ({
   is_test_bank: isTestInstitution,
   keywords: filterJunk(keywords),
   name,
-  logo: logoUrl,
+  logo: isEmptyStr(logoUrl) ? null : logoUrl,
```
wesrisenmay-mx, line 43: "" is falsy, so you can just do logoUrl || null

wesrisenmay-mx (approved):

PR #76, opened 2024-10-16
Wes/add aggregator integration

wesrisenmay-mx:


apps/institution-service/openApi.json
```diff
@@ -379,6 +374,30 @@
                           }
                         }
                       ]
+                    },
```
tphalp, line 377: Regarding the openapi spec, there are a few errors, according to the swagger editor. 
image

tphalp: I'm not sure if those errors are all that important to fix, but wanted to just make a note.

wesrisenmay-mx: Fixed

tphalp (approved):

PR #77, opened 2024-10-17
Adds search and filtering to institutions endpoint

Craiting:


tphalp (approved):

apps/institution-service/cypress/e2e/institutionsFiltersAndSearch.cy.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  it("gets a list of active institutions with mx agg support, supports_history, has OAuth and 'Bank' in the name", () => {
+    getInstitutionsRequest({
+      integrationFieldFilter: ["isActive", "supportsHistory", "supportsOauth"],
+      aggregatorFilter: ["mx"],
+      search: "Bank",
+    }).then(
+      (response: Cypress.Response<{ institutions: InstitutionDetail[] }>) => {
+        response.body.institutions.forEach((institution) => {
+          expect(institution.name).to.include("Bank");
```
wesrisenmay-mx, line 161: Is the search case sensitive?

apps/institution-service/openApi.json
```diff
@@ -482,6 +482,104 @@
...
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "required": false,
+            "style": "form",
+            "explode": true,
+            "example": ["mx", "sophtron"],
+            "description": "Only show Institutions that support extended transaction history"
```
wesrisenmay-mx, line 582: This description looks wrong

apps/institution-service/src/controllers/institutionController.ts
```diff
@@ -180,11 +190,80 @@ const getPaginationOptions = (req: Request): PaginationOptions => {
...
+
+const whereInstitutionConditions = (req: Request): WhereConditions => {
+  const { search, aggregatorName } = req.query;
+
+  const whereConditions: WhereConditions = {};
+
+  if (search) {
+    whereConditions[Op.or] = [
+      { name: { [Op.iLike]: `%${search as string}%` } },
+      { keywords: { [Op.contains]: [search as string] } },
```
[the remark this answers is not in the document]
Craiting: It's a bit trickier to do LIKE on an array but it's doable. It might need to be done in raw sql. This is currently case sensitive and has to be an exact match so it's not the best. I'm probably going to have to rewrite most of the sql query so I'll see if I can fix this while I'm fixing everything else.

apps/institution-service/cypress/e2e/institutionsFiltersAndSearch.cy.ts
Craiting: Yeah because the `iLike` operator is a case insensitive search

PR #81, opened 2024-10-29
Make some adjustments to institution filtering

Craiting:


apps/institution-service/cypress/e2e/institutionsFiltersAndSearch.cy.ts
```diff
@@ -185,4 +191,47 @@ describe("/institutions with filters and search", () => {
       },
     );
   });
+
+  it("includes institutions with inactive integrations when includeInactiveIntegrations is passed", () => {
+    getInstitutionsRequest({
+      aggregatorFilter: ["finicity"],
```
wesrisenmay-mx, line 197: I would make a test institution and filter by it's unique name so that this test will work long term

apps/institution-service/src/controllers/institutionController.ts
```diff
@@ -277,31 +310,35 @@ export const getPaginatedInstitutions = async (req: Request, res: Response) => {
...
-      distinct: true,
-      limit,
-      offset,
       order: [
         ["createdAt", "DESC"],
         ["name", "ASC"],
       ],
     });
 
+    const count = await Institution.count({
```
wesrisenmay-mx, line 330: Need updated integration tests for the changes in this file

Craiting: The previous integration tests already cover everything that changed in how the query is built. Do you think I should add additional tests here to duplicate the edge case tests which are covered in the E2E tests or what other tests should I add?

wesrisenmay-mx: I think we're out of whack with the testing trophy on this feature currently.

From our testing ADR: "The purpose of our end to end tests will be to validate that the frontend is working with the backend properly, and that apis are working together properly. We don't need to test every edge case in our end to end tests, because using MSW allows us to test those edge cases in our integration tests."

Every test we can write using an integration test should be done via an integration test. E2E tests are to add extra confidence, ensure that things are working together properly, and tests things that can't be tested via an integration test.

If we only wrote integration tests on this feature, then we could break the API because our integration tests don't know if express is setup properly. If we test all the filtering via integration tests, then we know if the correct parameters are passed, and express is setup properly, then it will continue to work. Adding an e2e test that really is just ensuring that the pieces are put together properly gives us the confidence we need.

If it were me I'd transfer a lot of the logic that the e2e tests are testing into integration tests, and just keep 1 or 2 e2e tests that test several filters all at once

apps/institution-service/src/test/institutionSearchAndFilter.test.ts
```diff
@@ -0,0 +1,332 @@
+/* eslint-disable @typescript-eslint/unbound-method */
+import { Request, Response } from "express";
```
wesrisenmay-mx, line 2: These tests should be in institutionController.test.ts

apps/institution-service/src/test/institutionSearchAndFilter.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    } as unknown as Response;
+
+    await getPaginatedInstitutions(req, res);
+
+    const jsonResponse =
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      (res.json as jest.Mock).mock.calls[0][0] as PaginatedInstitutionResponse;
+    expect(res.status).toHaveBeenCalledWith(200);
+    expect(jsonResponse.totalRecords).toBeGreaterThan(0);
+    jsonResponse.institutions.forEach((institution) => {
```
wesrisenmay-mx, line 120: I'd use .some

apps/institution-service/src/test/institutionSearchAndFilter.test.ts
```diff
@@ -0,0 +1,332 @@
...
+          aggInt.supports_history
+        ) {
+          jobTypesSupported = true;
+        }
+      });
+      expect(jobTypesSupported).toBeTruthy();
+    });
+  });
+
+  it("institution list by default only includes institutions which have at least one active integration", async () => {
```
wesrisenmay-mx, line 136: I think we want institutions to show up even if they don't have an active integration. When we're filtering by an aggregator, then it would need be an active integration for that aggregator.

It would be confusing to create a new institution, and have it not show up in the list until you click the includeActiveIntegrations switch

apps/institution-service/src/test/institutionSearchAndFilter.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    expect(res.status).toHaveBeenCalledWith(200);
+    expect(jsonResponse.totalRecords).toBeGreaterThan(0);
+    jsonResponse.institutions.forEach((institution) => {
+      expect(institution.name.toLowerCase().includes("wells")).toBeTruthy();
+    });
+  });
+
+  it("includes institutions with inactive integrations when includeInactiveIntegrations is passed", async () => {
+    const req = buildInstitutionRequest({
```
wesrisenmay-mx, line 272: You'll need to pass an aggregator to run this test

apps/institution-service/src/controllers/institutionController.ts
wesrisenmay-mx: Do we still have 1 e2e test to give us confidence that express is setup correctly for filtering?

apps/institution-service/openApi.json
```diff
@@ -496,15 +496,15 @@
...
+            "name": "includeInactiveIntegrations",
             "schema": {
               "type": "boolean",
               "enum": [true],
               "example": true
             },
             "required": false,
             "example": true,
-            "description": "Only show Institutions that are active"
+            "description": "Include institutions without active integrations in the search"
```
wesrisenmay-mx, line 507: When using aggregator integration filters include institutions that have inactive integrations that match the filters

apps/institution-service/src/controllers/institutionController.ts
Craiting: Ok I added one back to the institutions.cy file

wesrisenmay-mx (approved):

PR #82, opened 2024-11-06
Wes/institution search and filter UI

wesrisenmay-mx:


apps/ucp-ui/src/Institutions/InstitutionFilters.tsx
```diff
@@ -0,0 +1,183 @@
...
+          }
+          label={INSTITUTIONS_FILTER_INCLUDE_INACTIVE_INTEGRATIONS_LABEL_TEXT}
+          slotProps={slotProps}
+        />
+      </FormGroup>
+    </div>
+  );
+};
+
+export default InstitutionFilters;
```
tphalp, line 183: We previously decided to export inline, and not at the end of the file. Is that pattern different with Components?

wesrisenmay-mx: image

I just do it so I can name the the default export, and vscode will pick it up easier

tphalp: Makes sense. Thanks for explaining.

apps/ucp-ui/src/Institutions/InstitutionFilters.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    const [firstAggregator] = aggregators;
+
+    await userEvent.click(screen.getByLabelText(firstAggregator.displayName));
+
+    expectedAggregatorNames.splice(
+      expectedAggregatorNames.findIndex(
+        (name) => firstAggregator.name === name,
+      ),
```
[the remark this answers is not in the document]
wesrisenmay-mx: Good call. Fixed

Craiting (approved):

PR #84, opened 2024-11-15
Ct/heroku uncompiled

Craiting:
Staging urls:
InstitutionService: https://institution-service-staging-67817c56843f.herokuapp.com/
AuthenticationService: https://authentication-service-staging-1e08f5423f58.herokuapp.com/

### DIFF
--- a/apps/authentication-service/src/index.ts
+++ b/apps/authentication-service/src/index.ts
@@ -11,7 +11,8 @@ const app = express();
 
 initExpress(app);
 
-const listeningPort = process.env.PORT || PORT;
+const listeningPort =
+  process.env.NODE_ENV === "production" ? process.env.PORT || PORT : PORT;
 
 app.listen(listeningPort, () => {
   void getAccessToken()
--- a/apps/authentication-service/src/index.ts
+++ b/apps/authentication-service/src/index.ts
@@ -12,11 +12,13 @@ const app = express();
 
 initExpress(app);
 
-app.listen(PORT, () => {
+const listeningPort = process.env.PORT || PORT;
+
+app.listen(listeningPort, () => {
   void getAccessToken()
     .then(() => {
       console.log(
-        `\n${ConsoleColors.FgMagenta}${SERVICE_NAME} is listening on PORT ${PORT}; ENV=${envs.ENV}`,
+        `\n${ConsoleColors.FgMagenta}${SERVICE_NAME} is listening on PORT ${listeningPort}; ENV=${envs.ENV}`,
       );
       console.log(
         `${ConsoleColors.FgGreen}Service is initialized and ready to roll${ConsoleColors.Reset}`,
--- a/apps/authentication-service/src/index.ts
+++ b/apps/authentication-service/src/index.ts
@@ -11,8 +11,7 @@ const app = express();
 
 initExpress(app);
 
-const listeningPort =
-  process.env.NODE_ENV === "production" ? process.env.PORT || PORT : PORT;
+const listeningPort = process.env.PORT || PORT;
 
 app.listen(listeningPort, () => {
   void getAccessToken()
--- a/apps/authentication-service/src/index.ts
+++ b/apps/authentication-service/src/index.ts
@@ -1,10 +1,9 @@
 import express from "express";
 
 import envs from "./config";
-import { PORT } from "./shared/consts";
 import { initExpress } from "./init";
-import { getAccessToken } from "@/shared/auth/authService";
-import { ConsoleColors } from "@/shared/consts";
+import { getAccessToken } from "./shared/auth/authService";
+import { ConsoleColors, PORT } from "./shared/consts";
 
 export const SERVICE_NAME = "ucp-authentication-service";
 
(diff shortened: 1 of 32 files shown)


apps/authentication-service/src/index.ts
```diff
@@ -1,22 +1,24 @@
...
 
 export const SERVICE_NAME = "ucp-authentication-service";
 
 const app = express();
 
 initExpress(app);
 
-app.listen(PORT, () => {
+const listeningPort =
+  process.env.NODE_ENV === "production" ? process.env.PORT || PORT : PORT;
```
Craiting, line 15: Heroku assigns a dynamic port to listen on so this was necessary to get the port from the environment in heroku deployments.

tphalp (approved):

apps/authentication-service/src/index.ts
wesrisenmay-mx: Does process.env.PORT || PORT work?

apps/institution-service/package.json
```diff
@@ -20,12 +20,14 @@
     "test:db:seed": "npx sequelize-cli db:seed:all --env test",
     "test:db:reset": "npx sequelize-cli db:drop --env test && npm run test:db:setup",
     "dev": "nodemon ./src/app.ts",
+    "start": "npx ts-node --transpile-only ./src/app.ts",
```
wesrisenmay-mx, line 23: Why do we need a start command and a procfile? Is the start command just to test that the procfile command will work?

apps/institution-service/src/app.ts
```diff
@@ -81,8 +82,11 @@ app.use("/permissions", permissionsRoutes);
 app.use("/aggregatorIntegrations", aggregatorIntegrationRoutes);
 app.use("/aggregators", aggregatorRoutes);
 
-app.listen(PORT, () => {
-  console.info(`App listening on port ${PORT}`);
+const listeningPort =
+  process.env.NODE_ENV === "production" ? process.env.PORT || PORT : PORT;
```
wesrisenmay-mx, line 86: Same question here

package.json
```diff
@@ -11,19 +11,18 @@
   },
   "homepage": "https://github.com/Universal-Connect-Project/ucp-hosted-app#readme",
   "scripts": {
-    "build": "turbo build",
+    "build": "",
```
wesrisenmay-mx, line 14: Why remove this?

package.json
```diff
@@ -11,19 +11,18 @@
...
     "dev": "turbo dev",
     "lint": "turbo lint",
     "prettier": "turbo run prettier",
-    "test": "turbo run test",
-    "prepare": "husky"
```
wesrisenmay-mx, line 19: Why remove this?

apps/institution-service/src/database.ts
```diff
@@ -10,13 +10,19 @@ const config: { [key: string]: string } = require(
...
-    host: config.host,
-    port: config.port as unknown as number,
-  },
-);
+const sequelize = process.env.DATABASE_URL
+  ? new Sequelize(process.env.DATABASE_URL, {
+      dialectOptions: {
+        ssl: {
+          required: true,
+          rejectUnauthorized: false,
```
wesrisenmay-mx, line 18: why is rejectUnauthorized false?

apps/institution-service/package.json
Craiting: It's actually pointless in the package json, so I'll remove it. We need to use Procfile because the heroku apps allow an env variable to point to a specific procfile so that we can deploy different services on the same repo.

package.json
Craiting: I thought this was breaking the build process but just tested adding it back and it doesn't break so I'll leave it.

package.json
Craiting: Same here, initially I thought this was causing build problems but I just found out it's not the case.

apps/institution-service/src/database.ts
Craiting: When it's not false postgres wont allow the connection, I get this error in the heroku logs: 
`Error syncing database: ConnectionError [SequelizeConnectionError]: unable to get local issuer certificate`

I've tried enabling automatic certificate management on heroku and been looking for a way to run it with `rejectUnauthorized: true` but I've had no success yet. I'll keep looking.

It sounds like using `rejectUnauthorized: true` helps prevent man in the middle attacks, but since we don't have any sensitive data it might not matter. But if it's possible to get it to work then we might as well do it.

apps/authentication-service/src/index.ts
Craiting: That's what I did before but it messed up the pipelines because the PORT env variable wasn't set properly on those. I can probably change this code and then make sure the correct port is set in the environments for the github pipelines

apps/institution-service/src/database.ts
Craiting: From my research it seems like I can't set rejectUnauthorized to true because the `heroku postgres database` is a private database service which uses self-signed certificates which don't meet the standards for the public Certificate Authority. This is intentional because it gives heroku more flexibility to manage and update certificates internally without relying on the external Certificate Authority. So it sounds to me like we just kind of have to trust that heroku knows what it's doing with ssl. 

So we should still be getting encryption and security from heroku. From this research I feel like it's ok leave this setting as false. What do you think?

wesrisenmay-mx (approved):

PR #88, opened 2024-12-12
Institutions table sorting

tphalp:
Add ability to pass sort query in request for paginated institution api.

### DIFF
--- a/apps/institution-service/cypress/e2e/institutions.cy.ts
+++ b/apps/institution-service/cypress/e2e/institutions.cy.ts
@@ -17,6 +17,7 @@ import { createAuthorizationHeader } from "../shared/utils/authorization";
 import { getInstitutionsWithFiltersRequest } from "../shared/utils/institutions";
 import {
   createTestInstitution,
+  isSorted,
   runInvalidPermissionCheck,
   runTokenInvalidCheck,
 } from "../support/utils";
@@ -509,9 +510,13 @@ describe("/institutions", () => {
 
       expect(response.status).to.eq(200);
       expect(institutionResponse.totalRecords).to.eq(3);
-      expect(institutionResponse.institutions[0].id).to.eq(
-        "3b561893-e969-4a2c-9e58-3b81b203cdc1",
-      );
+      expect(
+        isSorted(
+          institutionResponse.institutions,
+          ["createdAt", "name"],
+          ["desc", "asc"],
+        ),
+      ).to.be.true;
     });
   });
 
@@ -528,9 +533,13 @@ describe("/institutions", () => {
 
       expect(response.status).to.eq(200);
       expect(institutionResponse.totalRecords).to.eq(3);
-      expect(institutionResponse.institutions[0].id).to.eq(
-        "7a909e62-98b6-4a34-8725-b2a6a63e830a",
-      );
+      expect(
+        isSorted(
+          institutionResponse.institutions,
+          ["id", "name"],
+          ["desc", "asc"],
+        ),
+      ).to.be.true;
     });
   });
 
--- a/apps/institution-service/cypress/e2e/institutions.cy.ts
+++ b/apps/institution-service/cypress/e2e/institutions.cy.ts
@@ -497,9 +497,9 @@ describe("/institutions", () => {
     );
   });
 
-  it("gets sorted institution list with default sort options", () => {
+  it.only("gets sorted institution list with default sort options", () => {
     cy.request({
-      url: `http://localhost:${PORT}/institutions?aggregatorName=testExampleA`,
+      url: `http://localhost:${PORT}/institutions`,
       method: "GET",
       headers: {
         Authorization: createAuthorizationHeader(SUPER_USER_ACCESS_TOKEN_ENV),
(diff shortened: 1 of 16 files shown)


apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -496,6 +496,63 @@ describe("/institutions", () => {
...
+
+      expect(response.status).to.eq(200);
+      expect(institutionResponse.totalRecords).to.eq(3);
+      expect(institutionResponse.institutions[0].id).to.eq(
+        "7a909e62-98b6-4a34-8725-b2a6a63e830a",
+      );
+    });
+  });
+
+  it("gets sorted institution list with multiple, custom sort options", () => {
```
wesrisenmay-mx, line 537: Why do we need multiple if there are only UI controls for 1?

tphalp: Good point about the UI. The default, hard-coded sort was for two columns, so I thought we needed to support that use-case.

tphalp: You can see that in the original code: 
image

apps/institution-service/src/controllers/institutionController.test.ts
```diff
@@ -707,6 +714,106 @@ describe("institutionController", () => {
 
       expect(inactiveInstitutionFound).toBeFalsy();
     });
+
+    it("calls getSortOption with default sort order if sortBy is not passed in to request", async () => {
```
wesrisenmay-mx, line 718: These tests are really mocky, so they're testing implementation details instead of actual behavior

https://kentcdodds.com/blog/testing-implementation-details

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -496,6 +496,63 @@ describe("/institutions", () => {
...
+      headers: {
+        Authorization: createAuthorizationHeader(SUPER_USER_ACCESS_TOKEN_ENV),
+      },
+    }).then((response: Cypress.Response<{ body: unknown }>) => {
+      const institutionResponse =
+        response.body as unknown as PaginatedInstitutionsResponse;
+
+      expect(response.status).to.eq(200);
+      expect(institutionResponse.totalRecords).to.eq(3);
+      expect(institutionResponse.institutions[0].id).to.eq(
```
wesrisenmay-mx, line 512: This is quite brittle. If someone creates another testExampleA bank, then this test could fail.

You could check the results of a request and make sure that the values are ordered correctly

apps/institution-service/src/controllers/institutionController.ts
```diff
@@ -191,9 +201,34 @@ const getPaginationOptions = (req: Request): PaginationOptions => {
...
 };
 
+const parseSort = (sortBy: string[]): SortSequelize[] => {
+  return sortBy.map((param) => {
+    const [column, direction = SortDirection.ASC] = param.split(":");
+    return { column, direction: direction as SortDirection };
+  });
+};
+
+export const getSortOption = (
```
wesrisenmay-mx, line 215: This looks like a premature abstraction to me. We don't have a second use case for this function, but it's built like it wants to be reused.

Abstracting is fine if it's only meant to be used by this one function, but then it should be built like that. DefaultSortString isn't necessary, because there's only one use case right now.

apps/institution-service/cypress/e2e/institutions.cy.ts
Craiting: I think that solution sounds good @wesrisenmay-mx. The reason we added the secondary sort was because the createdAt field was exactly the same for most institutions and it was messing up the pagination.

I agree with keeping the name as secondary sort for all other sorts. If they are sorting by name then you probably don't need a secondary sort, but making it sort by `id` seems like it wouldn't hurt anything.

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -17,6 +17,7 @@ import { createAuthorizationHeader } from "../shared/utils/authorization";
 import { getInstitutionsWithFiltersRequest } from "../shared/utils/institutions";
 import {
   createTestInstitution,
+  isSorted,
```
wesrisenmay-mx, line 20: isSorted sounds like it's a boolean instead of a function

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -496,6 +497,43 @@ describe("/institutions", () => {
...
+      method: "GET",
+      headers: {
+        Authorization: createAuthorizationHeader(SUPER_USER_ACCESS_TOKEN_ENV),
+      },
+    }).then((response: Cypress.Response<{ body: unknown }>) => {
+      const institutionResponse =
+        response.body as unknown as PaginatedInstitutionsResponse;
+
+      expect(response.status).to.eq(200);
+      expect(institutionResponse.totalRecords).to.eq(3);
```
wesrisenmay-mx, line 512: What's the purpose of checking that there are only 3 records? This will break if someone creates another institution with testExampleA in their local database. Also what's the purpose of filtering by testExampleA in this test?

apps/institution-service/cypress/e2e/institutions.cy.ts
```diff
@@ -496,6 +497,43 @@ describe("/institutions", () => {
...
+      expect(
+        isSorted(institutionResponse.institutions, ["createdAt"], ["desc"]),
+      ).to.be.true;
+    });
+  });
+
+  it("gets sorted institution list with a custom sort option", () => {
+    cy.request({
+      url: `http://localhost:${PORT}/institutions?aggregatorName=testExampleA&sortBy=id:desc`,
```
wesrisenmay-mx, line 521: is desc supposed to be upper case or lower case?

apps/institution-service/src/controllers/institutionController.test.ts
```diff
@@ -760,6 +764,60 @@ describe("institutionController", () => {
...
+      } as unknown as Response;
+
+      const findAndCountAllSpyCustom = jest.spyOn(
+        Institution,
+        "findAndCountAll",
+      );
+
+      await getPaginatedInstitutions(req, res);
+
+      expect(findAndCountAllSpyCustom).toHaveBeenCalledWith(
```
wesrisenmay-mx, line 785: Testing that findAndCountAll is called with a certain order is testing an implementation detail.

The sequelize api could change to be something other than order and these tests would still pass.

If you check the actual results of the function and test that they are being sorted correctly, then we have the confidence we need.

wesrisenmay-mx (approved):

PR #92, opened 2024-12-17
Changes to add sorting to institutions table in UI

tphalp:


apps/ucp-ui/cypress/e2e/institutions.cy.ts
```diff
@@ -102,6 +104,52 @@ describe("institutions", () => {
...
+
+    const rowRegex = new RegExp(INSTITUTIONS_ROW_TEST_ID);
+
+    cy.waitForLoad();
+
+    cy.findAllByTestId(rowRegex)
+      .eq(0)
+      .invoke("attr", "data-testid")
+      .then((testIdBeforeSort) => {
+        cy.get(INSTITUTIONS_TABLE_SORT_PARENT_CLASS)
```
wesrisenmay-mx, line 118: Can't you just click on the header label?

apps/ucp-ui/cypress/e2e/institutions.cy.ts
```diff
@@ -102,6 +104,52 @@ describe("institutions", () => {
...
+          .eq(0)
+          .find(INSTITUTIONS_TABLE_SORT_ARROW_CLASS_DOWN)
+          .eq(0)
+          .click();
+
+        cy.findAllByTestId(rowRegex);
```
wesrisenmay-mx, line 124: This doesn't seem to do anything

apps/ucp-ui/cypress/e2e/institutions.cy.ts
```diff
@@ -102,6 +104,52 @@ describe("institutions", () => {
...
+              .eq(0)
+              .click();
+
+            cy.waitForLoad();
+
+            cy.findAllByTestId(rowRegex)
+              .eq(0)
+              .invoke("attr", "data-testid")
+              .then((testIdFinalSort) => {
+                cy.waitForLoad();
```
wesrisenmay-mx, line 146: Why do you need to waitForLoad here?

apps/ucp-ui/cypress/e2e/institutions.cy.ts
tphalp: This is mainly my inexperience with cypress/testing-library. I'm open to simplified ways to do this. I got this idea from looking at other test files.

apps/ucp-ui/cypress/e2e/institutions.cy.ts
tphalp: Again, mainly my inexperience with this. I copied this from another test.

apps/ucp-ui/src/Institutions/Institutions.tsx
```diff
@@ -114,6 +125,14 @@ const Institutions = () => {
   );
   const search = searchParams.get("search") || "";
 
+  const [sortBy, setSortBy] = useState(
+    searchParams.get("sortBy") || `createdAt:${SortOrder.desc}`,
+  );
+  const [sortByProp, setSortByProp] = useState(sortBy.split(":")[0]);
+  const [sortByOrder, setSortByOrder] = useState<SortOrderType>(
```
wesrisenmay-mx, line 132: Can'y you just use SortOrder here instead of SortOrderType?

apps/ucp-ui/src/Institutions/Institutions.tsx
```diff
@@ -90,7 +101,7 @@ const Institutions = () => {
   const getBooleanFromSearchParams = (key: string) => {
     const value = searchParams.get(key);
 
-    return value === "true" ? true : false;
+    return value === "true";
```
wesrisenmay-mx, line 104: 👍

apps/ucp-ui/src/Institutions/Institutions.tsx
```diff
@@ -114,6 +125,14 @@ const Institutions = () => {
   );
   const search = searchParams.get("search") || "";
 
+  const [sortBy, setSortBy] = useState(
```
wesrisenmay-mx, line 128: It seems like we could just store one sortBy value in state, and derive these other sortBy variables from the thing stored in state

apps/ucp-ui/cypress/e2e/institutions.cy.ts
tphalp: I'll work on simplifying these selectors.

apps/ucp-ui/src/Institutions/Institutions.tsx
```diff
@@ -177,6 +197,23 @@ const Institutions = () => {
     });
   };
 
+  const handleChangeSortBy = (id: string) => () => {
```
wesrisenmay-mx, line 200: I'd call this createSortHandler

apps/ucp-ui/src/Institutions/Institutions.test.tsx
```diff
@@ -303,4 +308,95 @@ describe("<Institutions />", () => {
...
+      ),
+    );
+
+    await userEvent.click(
+      screen.getByText(INSTITUTIONS_TABLE_INSTITUTION_HEADER_TITLE),
+    );
+
+    await waitForLoad();
+
+    expect(
```
wesrisenmay-mx, line 379: I'd just search .within() the row to see if the id is there instead of worrying about where exactly it is in the table

apps/ucp-ui/src/Institutions/Institutions.test.tsx
```diff
@@ -303,4 +308,95 @@ describe("<Institutions />", () => {
...
+      container?.querySelector(INSTITUTIONS_TABLE_ROW_ROOT_CLASS)?.children[1], // Second column
+    ).toHaveTextContent(testSortList[0].id);
+
+    server.use(
```
wesrisenmay-mx, line 383: This tells us that it's calling the server again and looking for new results, but it doesn't really tell us that the sort parameters are integrated correctly.

I would setup the handler one time above, check the query parameters for sorting, and return the institutions with a different sort based on the value you read from the query parameters.

apps/ucp-ui/src/Institutions/Institutions.test.tsx
```diff
@@ -303,4 +308,95 @@ describe("<Institutions />", () => {
...
+      ),
+    );
+
+    await userEvent.click(
+      screen.getByText(INSTITUTIONS_TABLE_INSTITUTION_HEADER_TITLE),
+    );
+
+    await waitForLoad();
+
+    expect(
```
wesrisenmay-mx, line 398: I'd use within() getByText() here as well

apps/ucp-ui/src/Institutions/Institutions.test.tsx
tphalp: If we're testing sorting, doesn't it matter where it is?

wesrisenmay-mx (approved):

PR #106, opened 2025-01-15
Setup for staging and prod heroku envs

Craiting:


apps/ucp-ui/src/shared/constants/environment.test.ts
```diff
@@ -1,23 +1,23 @@
 describe("Environment based urls", () => {
   beforeEach(() => {
-    delete process.env.UCP_UI_REVIEW_APP;
+    delete process.env.REVIEW_APP;
```
Craiting, line 3: I decided to refactor and simplify this name because although it only applies to ucp-ui currently, it could be applied to any service. Also, because build is now checking for Review, staging, and production I thought removing the `UCP_UI` from the name simplifies things and makes it more readable.

build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
Craiting, line 1: Heroku generically uses the `build` script during the build phase. This is my attempt to cleanly run different builds for different heroku environments based on the environment variables set specific to each heroku environment. Different builds are necessary for different environment variables which need to be setup in the build phase for ucp-ui.

package.json
```diff
@@ -11,12 +11,14 @@
   },
   "homepage": "https://github.com/Universal-Connect-Project/ucp-hosted-app#readme",
   "scripts": {
-    "build": "if [ -z \"$UCP_UI_REVIEW_APP\" ]; then npm run build:main; else npm run build:review; fi",
+    "build": "./build.sh",
     "build:review": "npm run heroku:review:setupEnv && turbo build",
-    "build:main": "turbo build",
-    "build:prod": "turbo build:prod",
+    "build:staging": "npm run heroku:staging:setupEnv && turbo build",
```
Craiting, line 16: I didn't add this to turbo because it does what I need it to do with `turbo build`. The thing that's different is the environment variables set the institution and authentication service URLS to point to our heroku deployed services.

apps/ucp-ui/server.js
```diff
@@ -0,0 +1,22 @@
+import express from "express";
```
Craiting, line 1: I wasn't sure about using express to serve the front-end code, but chatgpt and a [stackOverflow](https://stackoverflow.com/a/41902436/25231320) I found claim that it's the way to do it. And it seems to work fine.

wesrisenmay-mx (approved):

PR #115, opened 2025-01-22
Wes/preferences

wesrisenmay-mx:


apps/institution-service/src/models/widgetPreferences.ts
```diff
@@ -0,0 +1,43 @@
...
+    defaultAggregator: Association<WidgetPreferences, Aggregator>;
+  };
+}
+
+WidgetPreferences.init(
+  {
+    id: {
+      type: DataTypes.STRING,
+      unique: true,
+      primaryKey: true,
```
Craiting, line 29: It might need to be updated here too
```suggestion
      type: DataTypes.UUID,
      unique: true,
      primaryKey: true,
      defaultValue: DataTypes.UUIDV4,
```

PR #116, opened 2025-01-29
Setup event endpoints for performance service

Craiting:


apps/authentication-service/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@repo/shared-utils": "*",
+    "@repo/backend-utils": "*",
```
Craiting, line 20: I added this new repo package for backend specific utils because when I tried adding some things which are used across all the backend services, the front-end code complained.

For example my backend function requires this package `express-oauth2-jwt-bearer`, but if you try to install that package on the front-end code then it says you need a bunch of polyfills.

apps/performance-service/cypress/shared/utils/authorization.ts
```diff
@@ -0,0 +1,24 @@
+export const createAuthorizationHeader = (cypressEnv: string) =>
```
Craiting, line 1: I tried putting some of the cypress functions into the shared repo package but struggled to get it to work because using cypress variables like `Cypress`, `cy.`, and `it` don't really work outside of the cypress configured environment. I tried a few different things and then decided I was spending a lot of time on it so I just moved on and copied the code into the performance service.

apps/performance-service/cypress/e2e/general/events.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+  runTokenInvalidCheck,
+} from "../../shared/utils/authorization";
+
+describe("connection event endpoints", () => {
+  const connectionId = "MBR-123";
+
+  describe("/events/:connectionId/connectionStart", () => {
+    const eventStartUrl = `events/${connectionId}/connectionStart`;
+    const eventRequestBody = {
+      jobType: "aggregation",
```
wesrisenmay-mx, line 13: This doesn't match what the job type looks like in ucw-app

apps/performance-service/cypress/e2e/general/events.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+        }).then((response: { status: number; body: { error: string } }) => {
+          expect(response.status).to.eq(400);
+          expect(response.body).to.have.property("error");
+          expect(response.body.error).to.include("jobType");
+        });
+      });
+    });
+
+    it("succeeds when jobType is one of the allowed values", () => {
+      const validJobTypes = [
```
wesrisenmay-mx, line 77: Can we use a shared enum for these?

apps/performance-service/cypress/e2e/general/events.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+          body: invalidBody,
+        }).then((response: { status: number; body: { error: string } }) => {
+          expect(response.status).to.eq(400);
+          expect(response.body).to.have.property("error");
+          expect(response.body.error).to.include("jobType");
+        });
+      });
+    });
+
+    it("succeeds when jobType is one of the allowed values", () => {
```
wesrisenmay-mx, line 76: It would be a little clearer when something fails if we had an it statement for each of the job types like you did on line 35

apps/performance-service/cypress/e2e/general/events.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+  createAuthorizationHeader,
+  runTokenInvalidCheck,
+} from "../../shared/utils/authorization";
+
+describe("connection event endpoints", () => {
+  const connectionId = "MBR-123";
+
+  describe("/events/:connectionId/connectionStart", () => {
+    const eventStartUrl = `events/${connectionId}/connectionStart`;
+    const eventRequestBody = {
```
wesrisenmay-mx, line 12: All this job type stuff is technically unknown until we know what's happening with combo jobs right?

apps/performance-service/cypress/support/e2e.ts
```diff
@@ -15,6 +15,80 @@
...
+  DefaultPermissions,
+  UiClientPermissions,
+  UiUserPermissions,
+} from "@repo/shared-utils";
+import { JwtPayload } from "jsonwebtoken";
+import { WIDGET_ACCESS_TOKEN } from "../shared/constants/accessTokens";
 import "./commands";
 
-before(() => {});
+const authenticateAndStoreToken = ({
```
wesrisenmay-mx, line 28: There's a loooooot of unused stuff in this function

apps/performance-service/openApi.json
```diff
@@ -0,0 +1,298 @@
...
+        ],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "jobType": {
+                    "type": "string"
```
wesrisenmay-mx, line 51: There's some sort of type that allows a certain set of strings

apps/performance-service/openApi.json
```diff
@@ -0,0 +1,298 @@
...
+                  },
+                  "institutionId": {
+                    "type": "string"
+                  },
+                  "AggregatorId": {
```
wesrisenmay-mx, line 56: these should be camelCase right?

packages/backend-utils/src/validation.ts
```diff
@@ -0,0 +1,28 @@
...
+  audience: string | string[];
+  auth0Domain: string;
+}) =>
+  auth({
+    audience: audience,
+    issuerBaseURL: `https://${auth0Domain}`,
+    tokenSigningAlg: "RS256",
+  });
+
+export const validateRequestBody = (schema: ObjectSchema) => {
```
wesrisenmay-mx, line 18: Need unit tests for this

apps/performance-service/cypress/e2e/general/events.cy.ts
Craiting: I think those job types will be the same with combo jobs but when we do combo jobs we will probably remove "all" and add new fields to represent each combination like "aggregate|verification"

PR #117, opened 2025-02-06
Add Redis and track connection progress

Craiting:


apps/performance-service/cypress/e2e/general/eventLifeCycle.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+  startConnectionEventRequest,
+  unpauseConnectionEventRequest,
+} from "../../shared/utils/requests";
+
+describe("connection event life cycle", () => {
+  const connectionId = crypto.randomUUID();
+
+  it(
+    "creates an event, pauses it, unpauses it, and marks it successfull then checks that " +
+      "it has been removed it from Redis by the processing poller",
```
wesrisenmay-mx, line 13: it has been removed

apps/performance-service/src/controllers/eventController.ts
```diff
@@ -1,47 +1,123 @@
...
-      .json({ message: "TODO: implement update challenge in next ticket" });
+    const { connectionId } = req.params;
+    const eventObj = (await getEvent(connectionId)) as EventObject;
+    if (eventObj.pausedAt) {
+      res.status(200).json({
+        message: "Connection process was already paused. Nothing changed.",
+        event: eventObj,
+      });
+    } else {
+      eventObj.pausedAt = Date.now();
```
wesrisenmay-mx, line 65: It's probably worth grabbing this Date.now() at the top of the function so we don't add delays of pulling things from redis

apps/performance-service/src/controllers/eventController.ts
```diff
@@ -1,47 +1,123 @@
...
   try {
-    // TODO: interact with Redis to update event to track when user interaction is happening
-    // so we can remove time spent by the user when collecting the duration metric
-    res
-      .status(200)
-      .json({ message: "TODO: implement challenge resumed in next ticket" });
+    const { connectionId } = req.params;
+    const eventObj = (await getEvent(connectionId)) as EventObject;
+    if (eventObj?.pausedAt) {
+      const pauseDurationMilliseconds = Date.now() - eventObj.pausedAt;
```
wesrisenmay-mx, line 83: Date.now at the top of the function

apps/performance-service/src/controllers/eventController.test.ts
```diff
@@ -6,26 +8,66 @@ import {
...
+
+const connectionId = "MBR-123";
+const expectRedisSetCall = (mockCallIndex: number, valueCheck: jest.Expect) => {
+  const mockSetCallArgs = mockSet.mock.calls[mockCallIndex];
+  const redisKey = mockSetCallArgs[0]; // First argument (key)
+  const redisValue = JSON.parse(mockSetCallArgs[1] as string); // Second argument (parsed JSON)
+
+  expect(redisKey).toBe(`${EVENT_SUBDIRECTORY}:${connectionId}`);
+
+  expect(redisValue).toEqual(valueCheck);
```
wesrisenmay-mx, line 23: It would be less mocky to check the value that is currently stored in redis instead of checking if things were called in a certain order

apps/performance-service/src/controllers/eventController.test.ts
```diff
@@ -36,70 +78,240 @@ describe("eventController", () => {
...
+          connectionId,
+        },
+        body: {},
+      } as unknown as Request;
+      const res = {
+        status: jest.fn().mockReturnThis(),
+        json: jest.fn(),
+      } as unknown as Response;
+
+      await setEvent(connectionId, {});
```
wesrisenmay-mx, line 100: We would have more confidence in our tests and be testing less implementation details if we called the functions that we would normally call to get this into the state we're expecting it to be in. For this test We could call createStartEvent instead of setEvent.

I would recommend auditing all the tests in this file and changing the instances of setEvent to the necessary functions that would setup the redis state to where it should be

apps/performance-service/src/services/storageClient/redis.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        {},
+      );
+    });
+  });
+
+  describe("getEvent", () => {
+    it("calls get on the client with event subdirectory", async () => {
+      await getEvent("test");
+
+      expect(mockGet).toHaveBeenCalledWith(`${EVENT_SUBDIRECTORY}:test`);
```
wesrisenmay-mx, line 84: think it would be better to check that getEvent returns the correct value

apps/performance-service/src/services/storageClient/redis.ts
```diff
@@ -0,0 +1,97 @@
...
+  await set(`${EVENT_SUBDIRECTORY}:${key}`, value);
+};
+
+export const getEvent = async (key: string) => {
+  return await get(`${EVENT_SUBDIRECTORY}:${key}`);
+};
+
+export const beginPollAndProcessEvents = () => {
+  return setIntervalAsync(
+    async () => {
```
wesrisenmay-mx, line 49:
```suggestion
processEvents,
```

apps/performance-service/src/services/storageClient/redis.ts
```diff
@@ -0,0 +1,97 @@
...
+export const setEvent = async (key: string, value: object) => {
+  await set(`${EVENT_SUBDIRECTORY}:${key}`, value);
+};
+
+export const getEvent = async (key: string) => {
+  return await get(`${EVENT_SUBDIRECTORY}:${key}`);
+};
+
+export const beginPollAndProcessEvents = () => {
+  return setIntervalAsync(
```
wesrisenmay-mx, line 48: You may be able to just do setIntervalAsync(processEvents)

apps/performance-service/src/services/storageClient/redis.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      const poller = beginPollAndProcessEvents();
+
+      jest.advanceTimersByTime(
+        Number(process.env.POLL_INTERVAL_SECONDS) * 1000 + 1000,
+      );
+
+      await clearIntervalAsync(poller);
+
+      expect(mockKeys).toHaveBeenCalledWith(`${EVENT_SUBDIRECTORY}:*`);
```
wesrisenmay-mx, line 101: The no keys test case feels like it should live in processEvents.

This function's it should probably just be: "triggers the process function on an interval"

apps/performance-service/src/controllers/eventController.test.ts
```diff
@@ -36,70 +82,198 @@ describe("eventController", () => {
...
+      await createStartEvent(mockRequest, res);
+
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      expect(res.status).toHaveBeenCalledWith(200);
+      expect(res.json).toHaveBeenCalledWith(
+        expect.objectContaining({
+          message: "Connection event already started",
+        }),
+      );
+      expect(mockSet).toHaveBeenCalledTimes(1);
```
wesrisenmay-mx, line 108: Can we check the object from get instead?

apps/performance-service/src/controllers/eventController.test.ts
```diff
@@ -36,70 +82,198 @@ describe("eventController", () => {
...
+      expect(res.status).toHaveBeenCalledWith(200);
+      expect(res.json).toHaveBeenCalledWith({
+        message: "Connection was not paused. Nothing changed.",
+        event: expect.objectContaining({
+          connectionId,
+          startedAt: expect.any(Number),
+        }),
+      });
+
+      expect(mockSet).toHaveBeenCalledTimes(1);
```
wesrisenmay-mx, line 273: Check the object instead

wesrisenmay-mx (approved):

PR #118, opened 2025-02-18
Add influxDb and read/write data

Craiting:


.github/workflows/auth-e2e-client-tests.yml
```diff
@@ -23,24 +23,22 @@ jobs:
...
-          echo AUTH0_CLIENT_SECRET=${{ secrets.AUTH0_CLIENT_SECRET }} >> ${ENV_FILE_PATH}          
+          echo AUTH0_CLIENT_SECRET=${{ secrets.AUTH0_CLIENT_SECRET }} >> ${ENV_FILE_PATH}
 
       - name: Cypress run (Standard suite)
         uses: cypress-io/github-action@v6
         env:
-          CYPRESS_AUTH_PASSWORD_WITH_KEY_ROLES: ${{ secrets.AUTH_PASSWORD_WITH_KEY_ROLES }}
-          CYPRESS_AUTH_USERNAME_WITH_KEY_ROLES: ${{ vars.AUTH_USERNAME_WITH_KEY_ROLES }}
-          CYPRESS_AUTH_PASSWORD_WITHOUT_KEY_ROLES: ${{ secrets.AUTH_PASSWORD_WITHOUT_KEY_ROLES }}
-          CYPRESS_AUTH_USERNAME_WITHOUT_KEY_ROLES: ${{ vars.AUTH_USERNAME_WITHOUT_KEY_ROLES }}
```
Craiting, line 34: The authentication-service tests started failing and after looking into it I think it's because the variables were using Tyson's users which look like they've been deleted from Auth0. After figuring out how to fix it I realized that some variables were not used so I deleted them and the other ones had a vague name "key roles" so I switched it to something more clear "widget role"

apps/performance-service/src/controllers/metricsController.ts
```diff
@@ -0,0 +1,14 @@
+import { Request, Response } from "express";
+import { getAndTransformAllInstitutionMetrics } from "../services/influxDb";
+
+export const getPerformanceRoutingJson = async (
```
wesrisenmay-mx, line 4: Needs a test

apps/performance-service/src/services/influxDb.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+interface QueryJobMetrics {
+  successRate: Record<string, number>;
+  jobDuration: Record<string, number>;
+}
+
+type InstitutionMetrics = Record<string, Record<string, QueryJobMetrics>>;
+
+function transformAllInstitutionsToJson(data: unknown[]) {
```
wesrisenmay-mx, line 61: Can we give this a type?

apps/performance-service/src/services/influxDb.ts
```diff
@@ -0,0 +1,153 @@
...
+  const jsonOutput: InstitutionMetrics = {};
+
+  data.forEach((row) => {
+    const { institutionId, jobTypes, aggregatorId, successRate, jobDuration } =
+      row as {
+        institutionId: string;
+        jobTypes: string;
```
wesrisenmay-mx: Is jobTypes a comma separated string? Are we ensuring that the jobTypes are always alphabetical or something, or how do we ensure that jobTypes in a different order are stored the same?

wesrisenmay-mx: I see the sort now. Nevermind.

apps/performance-service/src/services/influxDb.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    userInteractionTime: 2000,
+  };
+
+  it("records correct duration and success metrics on successful event", async () => {
+    const mockWriteApi = {
+      writePoint: jest.fn(),
+      close: jest.fn().mockResolvedValue(undefined),
+    } as unknown as WriteApi;
+
+    jest.spyOn(influxDb, "createNewWriteApi").mockReturnValue(mockWriteApi);
```
wesrisenmay-mx, line 35: Did you look into a way to make these tests more realistic and less mocky? I'm thinking of how you used a test database for the institution service

apps/performance-service/src/controllers/metricsController.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const res = {
+      send: jest.fn(),
+    } as unknown as Response;
+
+    const testInstitutionId = `bank1-${crypto.randomUUID()}`;
+    const testInstitutionId2 = `bank2-${crypto.randomUUID()}`;
+
+    await createTestScenarioEvents(testInstitutionId, testInstitutionId2);
+    await wait(2000);
```
Craiting, line 24: I had to add a wait period between writing to influx and reading because it doesn't write fast enough to be available for the read if I don't add a wait.

apps/performance-service/src/shared/tests/utils.ts
```diff
@@ -1,2 +1,38 @@
+import { queryApi } from "../../services/influxDb";
+
 export const minutesAgo = (minutes: number): number =>
   Date.now() - minutes * 60 * 1000;
+
+export const wait = (ms: number) =>
+  new Promise((resolve) => setTimeout(resolve, ms));
+
+export async function getLatestDataPoint(
```
wesrisenmay-mx, line 9: This file needs tests. Looks good otherwise

wesrisenmay-mx (approved):

PR #123, opened 2025-03-10
Add rewards and balance job types

Craiting:


apps/ucp-ui/src/shared/constants/jobTypes.ts
```diff
@@ -29,4 +31,14 @@ export const supportsJobTypeMap: Record<string, JobType> = {
     displayName: "Verification",
     prop: "supports_verification",
   },
+  rewards: {
+    description: "Account Rewards",
+    displayName: "Rewards",
+    prop: "supports_rewards",
```
wesrisenmay-mx, line 37: IMO we should stick with camel case for new props even though it wouldn't match the rest of the object

wesrisenmay-mx (approved):

PR #126, opened 2025-03-13
now using the clientID from the access token when using the start event

wesrisenmay-mx:


Craiting (approved):

Craiting (dismissed):

apps/performance-service/src/middlewares/validationMiddleware.ts
```diff
@@ -18,13 +28,30 @@ export async function validateConnectionId(
   const { connectionId } = req.params;
 
   if (!connectionId) {
```
wesrisenmay-mx, line 30: It's not possible for there to be no connectionId right? Because it's a path parameter?

packages/backend-utils/src/accessToken.ts
```diff
@@ -0,0 +1,5 @@
+import { Request } from "express";
+
+export const getAccessTokenFromRequest = (req: Request) => {
```
wesrisenmay-mx, line 3: Needs a test

wesrisenmay-mx:
We need e2e tests for the unauthorized clientId endpoint stuff. Because they are attached as middleware we don't have a way to test that with integration tests.

packages/backend-utils/src/accessToken.ts
Craiting: This gets tested indirectly in the [eventController.test.ts](https://github.com/Universal-Connect-Project/ucp-hosted-app/pull/126/files#diff-8ccf0f3c0e353958b0ea2f5360421130b4930d6d96443a212c455f2d839db9e9R22) tests and in the [validationMiddleware.test.ts](https://github.com/Universal-Connect-Project/ucp-hosted-app/pull/126/files#diff-70fd78da5ce4017fc738c4d94860fda252b22f816e3bae2b127f7f7d0c88b5f7R72) tests and in the eventLifeCycle.cy.ts, is that not enough?

Craiting:
> We need e2e tests for the unauthorized clientId endpoint stuff. Because they are attached as middleware we don't have a way to test that with integration tests.

The eventLifeCycle.cy test hits those endpoints proving that the middleware works. We don't have an e2e test that covers the failure scenario, but knowing that the success scenario works in e2e and the failure scenario works in the unit test I thought was enough proof. The only line that's not tested through the e2e test is line 54 but it is tested through the unit tests.
image

I feel like it's not necessary to add a failure scenario e2e test but what do you think after my arguments @wesrisenmay-mx ?

wesrisenmay-mx:
I forgot I made this PR, so it won't let me approve it, but I'm giving it the verbal approval so you can approve it yourself @Craiting

Craiting (approved):

PR #129, opened 2025-03-31
Add Aggregator Metrics data to performance service

Craiting:
Adds an endpoint to the performance service that delivers data necessary to create this page:
image


emilysallstrom (approved):
Nice work!

apps/performance-service/cypress/e2e/general/aggregatorMetrics.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+    cy.wait(6000); // 6 seconds for Redis processing poller to process and create a datapoint in Influx
+  });
+
+  it("fails on improper authorization", () => {
+    cy.request({
+      url: "metrics/aggregators",
+      method: "GET",
+      failOnStatusCode: false,
+      headers: {
+        Authorization: createAuthorizationHeader(UCP_UI_USER_ACCESS_TOKEN),
```
wesrisenmay-mx, line 31: Am I crazy or is this actually the authorization that should work?

@Craiting

apps/performance-service/openApi.json
```diff
@@ -21,6 +21,237 @@
...
+            "description": "Success",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "example": {
+                    "mx": {
+                      "avgSuccessRate": 0.4999163459929731,
+                      "avgDuration": 10928.8940940271,
+                      "jobTypes": [
```
wesrisenmay-mx, line 55: This data format makes it a bit difficult to render. The frontend will be rendered row by row, but this data is by aggregator which is the column.

apps/performance-service/src/shared/consts/index.ts
```diff
@@ -0,0 +1,7 @@
+export const TimeFrameAggWindowMap = {
```
wesrisenmay-mx, line 1: I had to think about this for about 10 seconds to realize what this was. Is there a more clear name you can think of?

TimeFrameToTimeBucketMap maybe?

apps/performance-service/cypress/e2e/general/aggregatorMetrics.cy.ts
wesrisenmay-mx: Ok I think I understand what's going on now. This endpoint is being called by the institution service and not the UI directly.

apps/performance-service/openApi.json
wesrisenmay-mx: This might be a moot point. Need to check the other PR to be sure.

apps/performance-service/src/aggregatorMetrics/aggregatorMetricsHandlers.ts
```diff
@@ -0,0 +1,130 @@
...
+    const data = await queryAndTransformAggregatorData(
+      (req.query?.timeFrame || "30d") as TimeFrame,
+    );
+    res.send(data);
+  } catch (error) {
+    res.status(400).json({ error });
+  }
+};
+
+const queryAndTransformAggregatorData = async (timeFrame: TimeFrame) => {
```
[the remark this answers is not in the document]
wesrisenmay-mx: This might be a moot point based on the other PR.

apps/performance-service/src/shared/consts/index.ts
Craiting: In flux it's called the aggregate window. I can change it to TimeFrameToAggregateWindowMap

PR #130, opened 2025-04-02
Include performance metrics with aggregators response (institution service)

Craiting:
Added endpoint to deliver data necessary to build this page:
image


emilysallstrom (approved):

emilysallstrom (approved):

apps/institution-service/openApi.json
```diff
@@ -35,35 +35,101 @@
...
-                        "updatedAt": "2024-10-09T17:01:07.587Z"
-                      },
-                      {
-                        "id": 98,
-                        "name": "mx",
-                        "displayName": "MX",
-                        "logo": "https://content.moneydesktop.com/storage/MD_Assets/Ipad%20Logos/100x100/INS-3aeb38da-26e4-3818-e0fa-673315ab7754_100x100.png",
-                        "createdAt": "2024-10-09T17:01:07.537Z",
-                        "updatedAt": "2024-10-09T17:01:07.537Z"
+                  "oneOf": [
```
wesrisenmay-mx, line 38: Adding this to the aggregators endpoint is going to increase the load time of the institutions page quite a bit. We render checkboxes for each aggregator, and adding performance data would mean that we have to wait for all the performance data before we can render those checkboxes.

apps/institution-service/openApi.json
```diff
@@ -35,35 +35,101 @@
...
+                          {
+                            "id": 98,
+                            "name": "mx",
+                            "displayName": "MX",
+                            "logo": "https://content.moneydesktop.com/storage/MD_Assets/Ipad%20Logos/100x100/INS-3aeb38da-26e4-3818-e0fa-673315ab7754_100x100.png",
+                            "createdAt": "2024-10-09T17:01:07.537Z",
+                            "updatedAt": "2024-10-09T17:01:07.537Z"
+                          }
+                        ],
+                        "warning": "Performance data failed to fetch",
```
wesrisenmay-mx, line 128: Why not just fail the request if something about the request failed?

Craiting: I thought about doing that and I'm fine with doing that, it just means if the performance service is not responsive then the aggregators response will always fail. So if you're running it locally you'll need to run both together as well which isn't a big deal. I was just trying to be flexible but it doesn't really matter.

apps/institution-service/openApi.json
Craiting: I thought we decided we wanted the whole page to load as quickly as possible instead of partial data. Are you thinking now that we should make the requests separately?

PR #131, opened 2025-04-08
use new endpoint for aggregator performance

Craiting:


apps/institution-service/src/controllers/aggregatorController.test.ts
```diff
@@ -1,12 +1,47 @@
...
-import { getAggregators } from "./aggregatorController";
+import {
+  getAggregators,
+  getAggregatorsWithPerformance,
+} from "./aggregatorController";
 import { server } from "../test/testServer";
 import { http, HttpResponse } from "msw";
 import { AGGREGATOR_PERFORMANCE_URL } from "../test/handlers";
 
 describe("getAggregators", () => {
```
wesrisenmay-mx, line 12: Needs a failure case test

wesrisenmay-mx (approved):

PR #156, opened 2025-06-16
Wes/actual performance charts UI

wesrisenmay-mx:


apps/ucp-ui/cypress/e2e/performance.cy.ts
```diff
@@ -1,18 +1,34 @@
...
+    const percentDataRegex = /^\d+(?:\.\d+)?%$/;
+    cy.findAllByText(percentDataRegex).should("have.length.at.least", 1);
+
+    const durationDataRegex = /^\d+(?:\.\d+)?s$/;
+    cy.findAllByText(durationDataRegex).should("have.length.at.least", 1);
 
-    cy.findAllByText(/^\d+(?:\.\d+)?%$/).should("have.length.at.least", 1);
-    cy.findAllByText(/^\d+(?:\.\d+)?s$/).should("have.length.at.least", 1);
+    const percentAndDurationDataRegex =
+      /^\d+(?:\.\d+)?%\s*\|\s*\d+(?:\.\d+)?s$/;
```
Craiting, line 29: I don't think this regex will be valid after I make my change to allow `0% | -`

wesrisenmay-mx: I'm actually ok if it stays like this anyways. The test is more to make sure that we're getting some complete data

apps/ucp-ui/src/shared/test/testData/performanceGraphts.ts
```diff
@@ -0,0 +1,80 @@
+const aggregatorThatsOnlyInDurationGraph = "durationGraphAggregator1";
```
Craiting, line 1: This filename seems to have a typo. `Graphts`

Craiting (commented):
Looks good. Just found the one filename with a typo

PR #178, opened 2025-07-21
Performance service rate limiting

wesrisenmay-mx:


apps/institution-service/src/useRateLimiting.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useRateLimiting } from "./useRateLimiting";
+
+describe("useRateLimiting", () => {
+  beforeEach(() => {
+    delete process.env.DISABLE_RATE_LIMITING;
+  });
+
+  beforeEach(() => {
+    delete process.env.DISABLE_RATE_LIMITING;
+  });
```
Craiting, line 11: duplicates

PR #179, opened 2025-07-22
added connections tab and refactored

ash-wright123:


wesrisenmay-mx:
We should match the border styling of the Institutions table: https://github.com/Universal-Connect-Project/ucp-hosted-app/blob/e783c018b9c9717814695e3e045aa0aab61dfe91/apps/ucp-ui/src/Institutions/institutions.module.css#L13

We should also change the border styling of the widget configuration to match. We don't use the bottom shadow that it currently shows.

apps/ucp-ui/src/Demo/DemoTabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+    },
+    [dispatch, connectionGuid, institutionName],
+  );
+
+  useEffect(() => {
+    window.addEventListener("message", handleMessage);
+    return () => {
+      window.removeEventListener("message", handleMessage);
+    };
+  }, [handleMessage]);
```
wesrisenmay-mx, line 62: Won't this add a new event listener every time handleMessageChanges, which is every time connectionGuid or institutionName change?

wesrisenmay-mx:
The redux strategy seems more complex than it needs to be.

Right now DemoTabs has the post message listener.
When it sees institution selected it stores the guid and institution name in local state.
When it sees member connected it dispatches the addConnection action.

The addConnection action looks in its state for a previously set aggregator and jobTypes and combines them with the provided institution to add to the store.

ConnectConfiguration dispatches the aggregatorLabel and jobTypeLabels to the store onSubmit.

I think we could have one component do the listening for the messages and store all the local state necessary to be able to dispatch completed connections to the redux store. Then we don't need the setConnectionDetails action

wesrisenmay-mx:
The file structure is confusing to me.

I would expect the file name that matches the folder to be the one that is the outermost container, so Demo.tsx instead of DemoTabs.tsx.

If I were to design the file structure it would look like this:

WidgetDemo/

WidgetDemo.tsx - renders the PageContent, the tabs, and the correct components for the tabs

Connect.tsx - renders the WidgetConfiguration form or the ConnectWidget. This can call useForm and pass what is necessary to WidgetConfiguration and ConnectWidget. This can store the state necessary for dispatching connections to the store and listen for postMessages.

WidgetConfiguration - receives necessary useForm props and renders the form

ConnectWidget - fetches the token, renders the error, loading state, children inside the PhoneContainer, renders the reset button

PhoneContainer - Generic phone container that just renders the phone wrapper + children

Connections - remains the same

apps/ucp-ui/src/Demo/DemoTabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+      if (type === INSTITUTION_SELECTED && metadata?.guid) {
+        setConnectionGuid(metadata.guid);
+        setInstitutionName(metadata.name || "");
+      }
+
+      if (
+        type === MEMBER_CONNECTED &&
+        metadata?.ucpInstitutionId &&
+        metadata.ucpInstitutionId === connectionGuid
```
wesrisenmay-mx, line 48: It might be safe to not check whether the ucpInstitutionId matches and just assume a connection success should use the last institution name. Can you think of a way that we could trick it into not matching?

apps/ucp-ui/src/WidgetDemo/Connect.tsx
```diff
@@ -0,0 +1,118 @@
...
+  });
+  const tokenErrorComponent = (
+    <FetchError
+      description={WIDGET_DEMO_ERROR_MESSAGE}
+      refetch={() => void refetch()}
+    />
+  );
+
+  return (
+    <PageContent>
```
wesrisenmay-mx, line 97: Page content is only meant to be rendered once per page. It looks like you are rendering a PageContent in WidgetDemo as well as a PageContent here

apps/ucp-ui/src/WidgetDemo/Connect.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <Stack spacing={4}>
```
wesrisenmay-mx, line 98: If there's only 1 thing in the stack, then this spacing doesn't do anything

apps/ucp-ui/src/WidgetDemo/PhoneContainer.tsx
```diff
@@ -5,7 +5,7 @@ import { SkeletonIfLoading } from "../shared/components/Skeleton";
 import { RESET_BUTTON_TEXT } from "./constants";
 
 interface PhoneContainerProps {
-  src: string;
+  src: React.ReactNode;
```
wesrisenmay-mx, line 8: It doesn't make sense to me to have src be either children or an iframe src url

ash-wright123: would you pass in two separate props?

wesrisenmay-mx:
If I were to use the Widget Demo UI and then look at the file structure it wouldn't be clear to me what is rendering what.

image

Connect matches the tab on the page, so I would think that Connect was what rendered the content for the Connect Tab.

ConnectConfiguration makes me think it's probably where the widget is configured, but it's less clear than WidgetConfiguration would be.

I wouldn't know where to find the component that renders the actual ConnectWidget. If I had used the UI enough then maybe I can surmise that it's in PhoneScreen.

IMO our file names generally make it clear what their job is. This structure still doesn't make it clear what the job of the component is.

apps/ucp-ui/src/WidgetDemo/connectWidget.module.css
```diff
@@ -0,0 +1,30 @@
...
+  box-shadow: 0px 0px 2px 0px rgba(0, 0, 0, 0.04);
+}
+.buttonContainer {
+  margin-top: 32px;
+  display: flex;
+  justify-content: center;
+}
+.errorContent {
+  padding-top: 25px;
+  height: 787px;
```
wesrisenmay-mx, line 29: Is there a different way we could style this that isn't a specific number of pixels that will adjust to the available height?

apps/ucp-ui/src/WidgetDemo/Connect.tsx
```diff
@@ -0,0 +1,163 @@
...
+  };
+
+  const handleReset = () => {
+    reset(defaultValues);
+    setSubmittedValues(null);
+  };
+
+  // POST MESSAGE HANDLER
+
+  const jobTypeDisplayNames = Object.entries(submittedValues || {})
```
wesrisenmay-mx, line 93: I'd probably move both of these into the useEffect, because they're only used there.

apps/ucp-ui/src/WidgetDemo/ConnectWidget.test.tsx
```diff
@@ -56,4 +58,16 @@ describe("<Demo />", () => {
     const iframe = await screen.findByTitle(WIDGET_DEMO_IFRAME_TITLE);
     expect(iframe).toBeInTheDocument();
   });
+  it("calls onReset when the reset button is clicked", async () => {
```
wesrisenmay-mx, line 61: Rendering the parent and checking that it goes back to rendering the the WidgetConfiguration would give more confidence than checking if a mock function was called.

apps/ucp-ui/src/WidgetDemo/PhoneContainer.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+describe("PhoneContainer", () => {
+  const mockProps = {
+    isLoading: false,
+  };
+
+  it("renders the skeletonLoader when isLoading is true", async () => {
+    render(
+      <PhoneContainer {...mockProps} isLoading={true}>
+        <iframe title="Test Iframe" src="https://example.com" />
```
wesrisenmay-mx, line 13: I'd just render text instead of an iframe

apps/ucp-ui/src/WidgetDemo/WidgetConfiguration.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      />,
+    );
+    await userEvent.click(
+      screen.getByRole("button", { name: LAUNCH_BUTTON_TEXT }),
+    );
+    expect(screen.getByText(`*${JOB_TYPE_ERROR_MESSAGE}`)).toBeInTheDocument();
+    expect(onSubmit).not.toHaveBeenCalled();
+  });
+
+  it("calls onSubmit with the correct data when form is valid", async () => {
```
wesrisenmay-mx, line 111: If we render the parent, then we can check to see if the iframe gets configured correctly instead of checking if a function was called in a particular way

apps/ucp-ui/src/WidgetDemo/WidgetConfiguration.tsx
```diff
@@ -0,0 +1,146 @@
...
+                <TextField
+                  fullWidth
+                  select
+                  label="Aggregator"
+                  data-testid="aggregator-select"
+                  {...field}
+                >
+                  {AGGREGATORS.map((aggregator) => (
+                    <MenuItem key={aggregator.value} value={aggregator.value}>
+                      {aggregator.label}
```
wesrisenmay-mx, line 124: You can use the aggregators from the endpoint to render the labels here

apps/ucp-ui/src/WidgetDemo/Connect.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    fireEvent(
+      window,
+      new MessageEvent("message", {
+        origin: WIDGET_DEMO_BASE_URL,
+        data: { type: MEMBER_CONNECTED },
+      }),
+    );
+
+    expect(dispatchSpy).toHaveBeenCalledWith(
```
wesrisenmay-mx, line 81: Instead of spying on the dispatch I would check the state of the store.

apps/ucp-ui/src/WidgetDemo/Connect.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    fireEvent(
+      window,
+      new MessageEvent("message", {
+        origin: "http://wrong-origin.com",
+        data: { type: MEMBER_CONNECTED },
+      }),
+    );
+
+    expect(dispatchSpy).not.toHaveBeenCalledWith(
```
wesrisenmay-mx, line 115: I would check the state instead of checking if dispatch is called here as well

apps/ucp-ui/src/WidgetDemo/constants.ts
```diff
@@ -2,7 +2,14 @@ export const WIDGET_DEMO_PAGE_TITLE = "Widget Demo";
...
 export const WIDGET_DEMO_IFRAME_TITLE = "Widget Demo Iframe";
 export const CONNECT_TAB = "Connect";
+export const CONNECTIONS_TAB = "Connections";
 export const RESET_BUTTON_TEXT = "Reset";
 export const LAUNCH_BUTTON_TEXT = "Launch";
 export const CONFIGURATION_HEADER = "Configuration";
 export const JOB_TYPE_ERROR_MESSAGE = "Please select at least one job type.";
+export const INSTITUTION_SELECTED = "connect/selectedInstitution";
+export const MEMBER_CONNECTED = "connect/memberConnected";
+export const AGGREGATORS = [
```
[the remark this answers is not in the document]
ash-wright123: But aren't there more aggregators than just mx and sophtron on that endpoint?

PR #189, opened 2025-08-26
adding plaid and akoya to initial institution db setup for developers

Craiting:


apps/institution-service/src/tasks/loadInstitutionsFromJson.ts
```diff
@@ -133,6 +148,40 @@ async function loadInstitutionData() {
         supports_aggregation: finicity.supports_aggregation,
       } as AggregatorIntegration);
     }
+
+    if (institution?.akoya?.id) {
+      const akoya = institution.akoya;
+      aggregatorList.push({
+        isActive: false,
```
wesrisenmay-mx, line 155: We would want Akoya to be active right?

PR #190, opened 2025-08-27
Add recordFailure flag to performance event object

Craiting:
Add recordFailure flag.

* once shouldRecordResult is set to `true` there is no setting it back to false
* if shouldRecordResult is true performance will be recorded


wesrisenmay-mx:
This looks good. Created this ticket as a followup: https://universalconnect.atlassian.net/jira/software/projects/UCP/boards/1?selectedIssue=UCP-753

If I'm blocked by this PR then we can merge before that followup ticket is done

wesrisenmay-mx (approved):

PR #201, opened 2025-09-17
optimize institution cache list with singleton pattern

Craiting:


apps/institution-service/src/performanceAuth/getPerformanceAuthInstitutions.test.ts
```diff
@@ -128,8 +128,11 @@ describe("getPerformanceAuthInstitutions", () => {
 
     const institutions = firstCall?.institutions as InstitutionDetail[];
 
+    expect(institutions.length).toBeGreaterThan(1);
     for (let i = 0; i < institutions.length - 1; i++) {
-      expect(institutions[i].name <= institutions[i + 1].name).toBeTruthy();
+      expect(
+        institutions[i].name.localeCompare(institutions[i + 1].name),
```
Craiting, line 134: This test was failing randomly and making this change seemed to fix it. Copilot says 

> Using localeCompare() handles locale-specific sorting rules and is more reliable across different environments.

wesrisenmay-mx (approved):

